PR #3, opened 2021-08-18
Apply many small fixes related to us now building on impish

sil2100:
Instead of focal/hirsute as we did previously. The snap now should build correctly.

To what should actually be part of core22 is like a separate discussion. For instance, the ubuntu-base tarball no longer ships bzip2 in impish by default, so I install it via extra packages - but maybe we won't need it in core22. Still, I'd rather start off with this as is and only later iterate.


hooks/990-ensure-uids-gids.chroot
```diff
@@ -28,10 +28,6 @@ echo "Ensure no user 107 (ex-docker)"
 cut -d: -f3 /etc/passwd | grep -q 107
 rc=$?; [ "$rc" = "0" ] && MISMATCH=1
 
-echo "Ensure no group 113 (ex-docker)"
```
stulluk, line 31: Why this has been removed ?

sil2100: There is a new group added in impish nowadays it seems, the `sgx` group - and it seems to use the 113 GID. So it feels to me that just erroring out on this here doesn't make sense, since it's like a completely new series.

stulluk (approved):
LGTM, thanks. Maybe we need to have a dedicated session to check if we really need all these packages anymore or not? 
@alfonsosanchezbeato FYI

alfonsosanchezbeato (approved):
LGTM, thanks!

PR #4, opened 2021-08-19
Fix snapcraft.yaml to build core22 instead of core20

sil2100:
This seems to have been reverted after the force push.

I am still not sure if we'll manage to get the builds working. It's hard for me to make launchpad use impish for the snap recipe build, and infering from snapcraft.yaml might not work. This is very annoying and very counterproductive so far.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,19 +1,20 @@
-name: core20
-# version: "20"
+name: core22
+# version: "22"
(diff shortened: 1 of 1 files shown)


stulluk (approved):
LGTM

alfonsosanchezbeato (approved):
LGTM, thanks

PR #6, opened 2021-09-02
Add bootchart configuration

alfonsosanchezbeato:
Add configuration files so it is possible to get a bootchart on
boot. The systemd-bootchart.service will not be enabled by default,
but it will be possible to enable it from kernel command
line (systemd.wants=systemd-bootchart.service) if we want it to run on
first boot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,3 +35,10 @@ stop when the `snapd.seeded.service` stops). The bootchart will be saved
 in the `ubuntu-save` partition, under `log/boot<N>/`, being `<N>` the
 boot number since bootcharts were enabled. If a chart has been collected
 by the initramfs, it will be also saved in that folder.
+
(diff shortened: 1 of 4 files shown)


alfonsosanchezbeato:
Same changes as for core20.

stulluk (approved):
LGTM, thanks. @alfonsosanchezbeato are there any changes that applied to other coreXX repos that are not yet available in this PR ?

alfonsosanchezbeato:
@stulluk @mvo this is now in sync with the bootchart implementation on core20 snap.

Meulengracht (approved):

alfonsosanchezbeato:
Merging as this is just syncing with what we have in the core20 repo (any further change would need to go to both repos anyway).

PR #8, opened 2021-09-28
fix libffi version for impish

stulluk:
in impish, libffi package name is simply libffi8
fixing this to let lp build correctly


alfonsosanchezbeato (approved):
LGTM, thanks

PR #9, opened 2021-10-04
Add a temporary hack to use the edge snapd for core22

sil2100:
This should fix the core22 builds on impish. We need to remember to remove this workaround once we have the current edge snapd version in stable.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -77,6 +77,11 @@ parts:
       - shellcheck
       - wget
       - distro-info
+    build-snaps:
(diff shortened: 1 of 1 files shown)


stulluk (approved):
LGTM

alfonsosanchezbeato (approved):
LGTM! Is there any LP bug to track the issue?

mvo5:
Yeah, that's fine, sorry for that, the fix should go to beta today/tomorrow but it will take a couple of days until it reaches stable :(

PR #12, opened 2021-11-03
Switch to jammy, and remove some core20 references.

sil2100:
This is mostly just a formality, since we already build using the current 'devel' tarball (as jammy is not out yet) - so we're pulling in the jammy tarball now anyway.

Note for the future: we might want to revisit some of the core20 hacks that are present in snapcraft.yaml and see if those are still relevant.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,8 @@ DPKG_ARCH := $(shell dpkg --print-architecture)
 #LTS=jj
 #BASE := $(LTS)-base-$(DPKG_ARCH).tar.gz
 #URL := http://cdimage.ubuntu.com/ubuntu-base/$(LTS)/daily/current/$(BASE)
-# Temporarily until JJ is open and released
-DEVEL=impish
+# Temporarily until JJ is released
+DEVEL=jammy
 BASE := $(DEVEL)-base-$(DPKG_ARCH).tar.gz
 URL := http://cdimage.ubuntu.com/ubuntu-base/daily/current/$(BASE)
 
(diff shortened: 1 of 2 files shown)


Meulengracht:
Looks good, LGTM.

Meulengracht (approved):

Meulengracht:
> LGTM thanks. I also managed to build core22 snap from your branch by this way: https://pastebin.ubuntu.com/p/xFRmcKrnGt/
> 
> Please note that, during snap install, I get a udev error, but then I installed snapcraft succesfully somehow (see line 309 in the above pastebin ) @mvo5 FYI ( Do I have to do anything about this? Or is it expected for non-stable tarballs ? )
I experience the exact same issue - I suspect it could be a container issue? Because I experience it on focal as well.

alfonsosanchezbeato:
LGTM! Note that it is possible to run locally in a confined container with something like
```
$ lxc launch images:ubuntu/jammy jammy
$ lxc config set jammy security.syscalls.intercept.mknod true
$ lxc shell jammy
# apt update
# apt install git snapd
$ exit
$ lxc restart jammy
# snap install --classic snapcraft
# git clone https://github.com/snapcore/core22.git -b formal-jammy
# cd formal-jammy
# snapcraft --destructive-mode
```
(the jey here is allowing mknod for some devices in the container, which is the same that `snapcraft` does for supported bases)

alfonsosanchezbeato (approved):

valentindavid:
> > Please note that, during snap install, I get a udev error, but then I installed snapcraft succesfully somehow (see line 309 in the above pastebin ) @mvo5 FYI ( Do I have to do anything about this? Or is it expected for non-stable tarballs ? )
> 
> I experience the exact same issue - I suspect it could be a container issue? Because I experience it on focal as well.

Is that https://bugs.launchpad.net/snapd/+bug/1712808 ?

PR #21, opened 2022-02-02
static: remove perl dependency

kubiko:
removing perl dependency from base snap

Signed-off-by: Ondrej Kubik <ondrej.kubik@canonical.com>

### DIFF
--- a/static/usr/bin/core-sshd-host-keygen
+++ b/static/usr/bin/core-sshd-host-keygen
@@ -9,10 +9,8 @@ get_config_option() {
 	[ -f /etc/ssh/sshd_config ] || return
 
 	# TODO: actually only one '=' allowed after option
-	perl -lne '
-		s/[[:space:]]+/ /g; s/[[:space:]]+$//;
-		print if s/^[[:space:]]*'"$option"'[[:space:]=]+//i' \
-	   /etc/ssh/sshd_config
+	awk -v option=$option 'tolower($1) == tolower(option) {print $2}' \
+		/etc/ssh/sshd_config
 }
 
 host_keys_required() {


valentindavid (approved):
Looks good to me.

To be noted however that keywords in `sshd_config` are case insensitive. So it should also report lines with "hOstKeY". The bug was already here. But if you manage to improve the matching before merging that would be nice.

I also wonder if the `host_keys_required` could have been simplified if we used `sshd -T` instead of `sshd_config` since it reports default values.

static/usr/bin/core-sshd-host-keygen
```diff
@@ -9,10 +9,8 @@ get_config_option() {
 	[ -f /etc/ssh/sshd_config ] || return
 
 	# TODO: actually only one '=' allowed after option
-	perl -lne '
-		s/[[:space:]]+/ /g; s/[[:space:]]+$//;
-		print if s/^[[:space:]]*'"$option"'[[:space:]=]+//i' \
-	   /etc/ssh/sshd_config
+	awk -v option=$option '$1 == option {print $2}' \
+        /etc/ssh/sshd_config
```
alfonsosanchezbeato, line 13:
```suggestion
	awk -v option=$option '$1 == option {print $2}' /etc/ssh/sshd_config
```

xnox:
My understanding was that snapd was at some point calling some tool, from whithin the base snap, that was available as perl only. Has that been fixed in snapd to move to the other tool which is written in C?

(this is just a check to ensure we have the other largish perl dep dropped to be able to remove perl from the base snap)

kubiko:
@valentindavid I have addressed now case insensitive  sshd config now as well

alfonsosanchezbeato (approved):
LGTM, thanks!

valentindavid (approved):

PR #25, opened 2022-02-09
Remove writable paths

valentindavid:
Instead we can use /etc/fstab and tmpfiles.d

This requires change in core-initrd to pre-run tmpfiles on
core-writable.conf before switch of root.

See snapcore/core-initrd#74


static/usr/lib/systemd/system/boot-uboot.mount
```diff
@@ -0,0 +1,10 @@
+[Unit]
+RequiresMountsFor=/run/mnt/ubuntu-boot
+ConditionPathExists=/run/mnt/ubuntu-boot/uboot/ubuntu/boot.sel
+ConditionKernelCommandLine=snapd_recovery_mode=run
+
+[Mount]
+Where=/boot/uboot
+What=/run/mnt/ubuntu-boot/uboot/ubuntu
+Type=none
+Options=bind
```
xnox, line 10: To support booting classic systems (like the cloudimg-rootfs feature) it would be nice if such units would be generated by snap-bootstrap itself, instead of being hardcoded in the core-initrd.

However, migrating such mounts to snap-bootstrap is a time consuming task.

static/etc/fstab
```diff
@@ -0,0 +1,49 @@
+/run/mnt/base / none bind,x-initrd.mount,ro 0 0
```
xnox, line 1: Previously the paths that are needed by a given base was encoded inside a gitven base, instead of initrd. Thus allowing changes to base paths as needed inside i.e. core22 to be self-contained inside core22 snap.

With this change it become more coupled, where for example a package change in core22 snap requires simultanious change in core-initrd.

Also consider that a given kernel (core-initrd) might be booting and rolling back across different core22 revisions in recovery mode & run mode.

This also makes it harder to boot classic systems, as most of these mounts must not be done when booting classic systems with for example cloudimg-rootfs feature.

Do you think we could keep this file inside core22 snap at all? I understand that it might be harder to change fstab dynammically, and change what it does.

However, I was under the impression that systemd in the initrd can read $(rootfs)/etc/fstab and do things as defined there, no? Meaning it should be possible to keep this inside core22 snap "by magic".

valentindavid: I am not sure I understand your comment.

This PR is for core-base. The corresponding PR for core-initrd in snapcore/core-initrd#74. I have removed from core-initrd as much as possible to move it to core-base. Now only `sysroot.mount` and `sysroot-writable.mount ` are left in core-initrd. `/usr/lib/modules`, `/usr/lib/firmware`, `/boot/*`are moved to core-base, where they should be.

So the coupling is actually reduced by this change.

> However, I was under the impression that systemd in the initrd can read $(rootfs)/etc/fstab and do things as defined there, no? Meaning it should be possible to keep this inside core22 snap "by magic".

This is exactly what snapcore/core-initrd#74 does.

xnox: I was confused indeed. I thought this PR was for core-initrd. This is great!

hooks/993-factory-files.chroot
```diff
@@ -0,0 +1,51 @@
...
+set -eu
+
+# cp -a requires /proc/self/fd in some cases
+mount -t proc proc /proc
+trap 'umount /proc' EXIT
+
+mkdir -p /usr/share/factory/writable/user-data
+cp -aT /home /usr/share/factory/writable/user-data/home
+
+dirs=(
```
alfonsosanchezbeato, line 14: iiuc this is these are the auto/persistent/transition entries from the old `writable-paths` file, and I see two misses:
```
/etc/default/crda
/var/lib/apparmor
```
I see that actually those folders do not exist in UC20, so probably this is fine, but I'd like to understand what happened with them, if somebody knows. Especially with crda which is actually a file in classic.

static/etc/fstab
```diff
@@ -0,0 +1,49 @@
...
+/run/mnt/data /writable none bind,x-initrd.mount 0 0
+/run/mnt/kernel/modules /usr/lib/modules none bind,x-initrd.mount,ro 0 0
+/run/mnt/kernel/firmware /usr/lib/firmware none bind,x-initrd.mount,ro 0 0
+
+/run/mnt/data/system-data/etc/systemd /etc/systemd none bind,x-initrd.mount 0 0
+/run/mnt/data/system-data/etc/writable /etc/writable none bind,x-initrd.mount 0 0
+/run/mnt/data/system-data/etc/dbus-1 /etc/dbus-1 none bind,x-initrd.mount 0 0
```
alfonsosanchezbeato, line 8: It would be good to have lines sorted in this file, does not seem to be fully the case

hooks/993-factory-files.chroot
alfonsosanchezbeato: And same for `/etc/apparmor.d/cache`, I guess.

hooks/993-factory-files.chroot
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+# tmpfiles.d will copy files from /usr/share/factory So file that
+# needs to be populated in /writable have to be copied there.
```
alfonsosanchezbeato, line 3: I think it would be good to add a more extensive comment mentioning `core-writable.conf` and a quick mention of what "C" and "d" options do, and `s/tmpfiles.d/tmpfiles.d(5)/` so it is clear that the man page can be consulted for details.

hooks/993-factory-files.chroot
```diff
@@ -0,0 +1,51 @@
...
+    /var/lib/systemd
+    /var/log
+    /var/snap
+    /var/tmp
+)
+
+for dir in "${dirs[@]}"; do
+  parent="$(dirname "${dir}")"
+  mkdir -p "/usr/share/factory/writable/system-data/${parent}"
+  cp -aT "${dir}" "/usr/share/factory/writable/system-data/${dir}"
```
alfonsosanchezbeato, line 50: How much does the base grow with these copies? Do we really need this? Reading tmpfiles.d(5) it seems that we can specify the source for where to copy files from

alfonsosanchezbeato (commented):
This is very nice. I have some questions around. Also, `README.md` should grow some instructions and what a developer would need to do if a new writable path needs to be added.

hooks/993-factory-files.chroot
valentindavid: mksquashfs detects duplicated files. The overhead should be marginal. Just directories.

Makefile
```diff
@@ -78,9 +78,6 @@ test:
 			if !(cd $(TESTDIR) && $$f); then \
 				exit 1; \
 			fi; \
-	    	done; \
-	set -ex; for f in $$(pwd)/tests/test_*.sh; do \
-		sh -e $$f; \
```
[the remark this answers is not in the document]
valentindavid: Done

static/etc/fstab
valentindavid: Done

valentindavid:
> Also, `README.md` should grow some instructions and what a developer would need to do if a new writable path needs to be added.

I have added that.

xnox:
hooray

README.md
alfonsosanchezbeato, line 1: Thanks for adding documentation

alfonsosanchezbeato (commented):
Thanks, mainly comments about fstab mounts

static/usr/lib/systemd/system/boot-efi.mount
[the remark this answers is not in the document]
valentindavid: This was removed. It was broken. I made a generator instead.

static/usr/lib/systemd/system/boot-piboot.mount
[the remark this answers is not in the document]
valentindavid: Same here.

static/etc/fstab
```diff
@@ -0,0 +1,54 @@
+/run/mnt/base / none bind,x-initrd.mount,ro 0 0
```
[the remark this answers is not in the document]
valentindavid: I will test removing that one while you re-review and see if `-.mount` is ok without it.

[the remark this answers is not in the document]
alfonsosanchezbeato: Thanks - if in the end the line is needed, it would be good to have a comment here. Also, I would prefer somthing like `/ / none bind,x-initrd.mount,ro 0 0` if possible.

static/etc/fstab
```diff
@@ -0,0 +1,54 @@
...
+/writable /writable none bind,x-initrd.mount 0 0
```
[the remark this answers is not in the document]
alfonsosanchezbeato: Ok, maybe good to have a comment too

static/etc/fstab
```diff
@@ -0,0 +1,54 @@
...
+/run/mnt/kernel/firmware /usr/lib/firmware none bind,x-initrd.mount,ro 0 0
+/run/mnt/kernel/modules /usr/lib/modules none bind,x-initrd.mount,ro 0 0
```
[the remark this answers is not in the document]
alfonsosanchezbeato: Ok, maybe worth a comment to clarify that the What will not match this in normal run mode

static/usr/lib/core/run-snapd-from-snap
alfonsosanchezbeato, line 1: debug left-over?

alfonsosanchezbeato (commented):
Thanks, some minor things

static/etc/fstab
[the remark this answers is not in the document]
valentindavid: I have tested with this one removed as well. Same thing. I will remove it.

Meulengracht (approved):
This looks super good, I'm a bit confused on the kernel-modules components and why it won't work. I'll ask for an explanation tomorrow

hooks/993-factory-files.chroot
```diff
@@ -0,0 +1,58 @@
...
+# that describes the copy of those files/directories.
+
+set -eu
+
+# cp -a requires /proc/self/fd in some cases
+mount -t proc proc /proc
+trap 'umount /proc' EXIT
+
+mkdir -p /usr/share/factory/writable/user-data
+cp -aT /home /usr/share/factory/writable/user-data
```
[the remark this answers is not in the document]
valentindavid: It should be empty. I can replace that by a check it is so.

pedronis (approved):
thanks

PR #28, opened 2022-02-15
spread: update to core22

Meulengracht:
Updated the tests to run on jammy, and ported a lot of the improvements from core-initrd repo to this repository.

I'll do a followup PR where I make sure that a subset of snapd tests also run


.github/workflows/lxd-image.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
           SPREAD_GOOGLE_KEY: ${{ secrets.SPREAD_GOOGLE_KEY }}
         run: |
           echo $SPREAD_GOOGLE_KEY > sa.json
-          spread google-nested:tests/spread/ci
+          spread google-nested:tests/spread/ci/
```
alfonsosanchezbeato, line 21: I'm curious, is the end slash really needed or just cosmetic?

tests/spread/ci/build-image/task.yaml
```diff
@@ -9,17 +9,15 @@ execute: |
...
     snap install lxd --channel=latest/stable
     lxd init --auto
 
     # authenticate with gcloud
     gcloud auth activate-service-account --key-file="./sa.json"
 
-    # launch the new lxc instance from an ubuntu 20.04 base
-    lxc launch ubuntu:20.04 $INSTANCE_NAME
+    # launch the new lxc instance from an ubuntu 22.04 base
+    lxc launch images:ubuntu/jammy $INSTANCE_NAME
```
alfonsosanchezbeato, line 20: I know we have to use the `ubuntu/jammy` image because `22.04` is not released yet, but I was wondering, what will be the difference when that happens? If not exactly the same we need to decide which one we would use after the release.

alfonsosanchezbeato (approved):
LGTM! I just have a couple of questions.

.github/workflows/lxd-image.yaml
Meulengracht: it's needed, otherwise spread will say no filter match

tests/spread/ci/build-image/task.yaml
Meulengracht: I unfortunately don't know this, I'm using ubuntu/jammy for the reason you said, but I don't know what will happen or if we will need to update this. Maybe someone else can answer this?

valentindavid (approved):

tests/spread/ci/build-image/task.yaml
Meulengracht: I added the TODO

PR #29, opened 2022-02-23
tests: support for snapd smoke-tests

Meulengracht:
This PR creates a new action workflow that executes snapd tests on both PRs and the main branch when comitted. With no more workarounds!

For now  only the smoke tests are executed, if we want other suites running then just tell me and I'll add them

### DIFF
--- a/.github/workflows/snapd-tests.yaml
+++ b/.github/workflows/snapd-tests.yaml
@@ -45,7 +45,7 @@ jobs:
           echo "************* DOWNLOADING SNAPS *************"
           download_core22_snaps 'edge'
 
-          echo "************* WRITING SSH CONFIG *************"
+          echo "************* WRITING CLOUD-INIT CONFIG *************"
           prepare_core22_cloudinit
 
           echo "************* BUILDING CORE22 SNAP *************"
--- a/.github/workflows/snapd-tests.yaml
+++ b/.github/workflows/snapd-tests.yaml
@@ -1,0 +1,69 @@
+name: Spread tests (snapd)
+on:
+  schedule:
+    # Weekly sunday testing for bitrot
+    - cron: '0 12 * * 0'
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Cleanup job workspace
+        id: cleanup-job-workspace
+        run: |
+          rm -rf "${{ github.workspace }}"
+          mkdir "${{ github.workspace }}"
+      
+      - uses: actions/checkout@v2
+        with:
+          path: core-base
+          ref: main
+      - uses: actions/checkout@v2
+        with:
+          repository: 'snapcore/snapd'
+          path: snapd
+          ref: master
+      
+      - name: Install spread
+        run: curl -s https://storage.googleapis.com/snapd-spread-tests/spread/spread-amd64.tar.gz | sudo tar xzv -C /usr/bin
+
+      - name: Build image
+        working-directory: '${{ github.workspace }}'
+        run: |
+          . "core-base/tests/lib/prepare-snapd.sh"
+
+          echo "************* INSTALLING DEPS *************"
+          install_core22_deps
+
+          echo "************* DOWNLOADING SNAPS *************"
+          download_core22_snaps 'edge'
+
+          echo "************* WRITING SSH CONFIG *************"
+          prepare_core22_cloudinit
+
+          echo "************* BUILDING CORE22 SNAP *************"
+          build_core22_snap '${{ github.workspace }}/core-base'
+
+          echo "************* BUILDING CORE22 IMAGE *************"
+          build_core22_image
+
+          echo "************* STARTING CORE22 VM *************"
+          start_snapd_core_vm '${{ github.workspace }}'
+
+          cd snapd
+          SPREAD_EXTERNAL_ADDRESS=localhost:8022 spread external:ubuntu-core-22-64:tests/smoke/
+
+      - name: Discard spread workers
+        if: always()
+        run: |
+          shopt -s nullglob
+          for r in .spread-reuse.*.yaml; do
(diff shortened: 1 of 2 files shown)


valentindavid (approved):

tests/lib/prepare-snapd.sh
```diff
@@ -0,0 +1,209 @@
...
+}
+
+build_core22_snap() {
+    local project_dir="$1"
+    local current_dir="$(pwd)"
+    
+    # run snapcraft
+    (
+        cd "$project_dir"
+        sudo snapcraft --destructive-mode
```
alfonsosanchezbeato, line 194: How do we know if we are in the right Ubuntu release to build in destructive mode? I think we need some check maybe by using lsb_release to avoid shooting us in the foot. I am also wondering where is the meaning of `runs-on: ubuntu-latest` defined.

tests/lib/prepare-snapd.sh
```diff
@@ -0,0 +1,209 @@
...
+users:
+  - name: ubuntu
+    sudo: "ALL=(ALL) NOPASSWD:ALL"
+    lock_passwd: false
+    plain_text_passwd: 'ubuntu'
+  - name: external
+    sudo: "ALL=(ALL) NOPASSWD:ALL"
+    lock_passwd: false
+    plain_text_passwd: 'ubuntu'
+  - name: test
```
alfonsosanchezbeato, line 161: Why do we need all these users? Is that because the tests require them? Some comment would be helpful.

tests/lib/prepare-snapd.sh
```diff
@@ -0,0 +1,209 @@
...
+    sudo: "ALL=(ALL) NOPASSWD:ALL"
+    lock_passwd: false
+    plain_text_passwd: 'ubuntu'
+    uid: "12345"
+
+bootcmd:
+  - mkdir -p /var/lib/console-conf
+  - touch /var/lib/console-conf/complete
```
alfonsosanchezbeato, line 169: This works, but the supported way afaiu is by setting a default in `gadget.yaml`:
```
defaults:
  system:
    service:
      console-conf:
        disable: true
```
(ignore this if the advice from snapd team is to prefer the boot commands)

.github/workflows/snapd-tests.yaml
```diff
@@ -0,0 +1,67 @@
...
+          build_core22_snap '${{ github.workspace }}/core-base'
+
+          echo "************* BUILDING CORE22 IMAGE *************"
+          build_core22_image
+
+          echo "************* STARTING CORE22 VM *************"
+          start_snapd_core_vm '${{ github.workspace }}'
+
+          cd snapd
+          SPREAD_EXTERNAL_ADDRESS=localhost:8022 spread external:ubuntu-core-22-64:tests/smoke/
```
alfonsosanchezbeato, line 59: Maybe it would be good to add a comment saying that here we could add more suites.

alfonsosanchezbeato (commented):
Looks great! I have some minor comments.

tests/lib/prepare-snapd.sh
Meulengracht: To know what ubuntu-latest refers to one can check the docs - however ubuntu-latest always points to the latest LTS release. So at this moment 20.04 - but it will switch to ubuntu 22.04 once it's released. Is it only a specific ubuntu release we can use? I can just change the runner to require a specific version.

The only reason I'm building in destructive mode is just for the speed of it, we are already executing in a VM, so no reason to spin a VM up to build in an isolated context. Unless I'm misunderstanding something?

tests/lib/prepare-snapd.sh
Meulengracht: Yea good point here - unfortunately all of them are required by various stages of testing. 'external' is used by the external backend, and 'test' is used during setup by snapd. Now that I think about it I'm not sure 'ubuntu' is required, I'll just try to verify if they are all required still, and then update the comments

tests/lib/prepare-snapd.sh
Meulengracht: Really good point, so strictly speaking I don't need to disable console-conf, which makes it weird why I have it here. I'll remove it, I was using it during testing to make it easier to navigate (and didn't really think about removing it again). Snapd team has not advised me to do this, but they suggest this solution on the forums as a way of disabling console-conf.

.github/workflows/snapd-tests.yaml
Meulengracht: Really good point, thanks for the review!

tests/lib/prepare-snapd.sh
alfonsosanchezbeato: The problem is that we should not use `--destructive-mode` if not on the same LTS as the base - we would be getting dependencies from the wrong place. I actually thought we were using 22.04 here and I was asking to check that was actually the case to avoid problems in the future if the version changed.

How is this done to build core-initrd in the tests? I assume we are using 22.04 in the main branch, is that not the case?

Meulengracht: Ah then I understand, in core-initrd and core-base (main tests) we actually spin up a google instance that runs ubuntu 22.04 and then run ```--destructive-mode``` there. So we use 22.04 as build base in those cases. This would be the only case not doing that, which is then wrong. 

It won't work without the --destructive-mode as the LXD base for 22.04 doesn't exist AFAIK?

tests/lib/prepare-snapd.sh
Meulengracht: I've updated this and added comments, I have removed 'ubuntu'

tests/lib/prepare-snapd.sh
Meulengracht: Ok, so unfortunately there is no easy way of doing this right now. I need to manually spin up a jammy instance, and then do ```--destructive-mode``` inside that, it seems like it's the way of doing this for core22 bases right now, as no LXD image exists for core22, only for jammy classic. I'll have to modify this to run in a jammy VM.

PR #32, opened 2022-02-25
tests: add support for executing snapd tests (v2)

Meulengracht:
This PR adds support for running snapd tests. The PR also introduces a seperation of build and executing tests to improve the test execution time. 

So instead of the snapd pipeline having to run nested to first build the snap in a jammy environment, we now build the snap first in a jammy environment, and then share that across build pipelines, removing the need for each test pipeline to build their own copy and removes the need for snapd tests to run nested.


valentindavid (approved):

alfonsosanchezbeato (approved):
LGTM! Really like how you run spread itself to build the base :-)

PR #35, opened 2022-03-15
Stop running auto-import directly from udevd

valentindavid:
`snap auto-import` cannot run anymore because of sandboxing of
systemd-udevd. Instead we need to schedule a systemd service.

~~It could be an issue for users using udisks2, where a logged in user
would automatically mount disks marked as `UDISKS_AUTO`. For a short
time, snap could have mounted the device and udisks might fail to
mount.  For this reason we introduce `SNAP_AUTO_IMPORT` so that users
can introduce a custom udev rule to disable auto import on devices
(all or filtered).~~

### DIFF
--- a/static/usr/lib/systemd/system/snapd.autoimport-device@.service
+++ b/static/usr/lib/systemd/system/snapd.autoimport-device@.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Auto import assertions from a specific block device
+After=snapd.service snapd.socket snapd.seeded.service
+ConditionKernelCommandLine=snapd_recovery_mode=run
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/snap auto-import --mount=/dev/%i
+PrivateMounts=yes
--- a/static/usr/lib/udev/rules.d/66-snapd-autoimport.rules
+++ b/static/usr/lib/udev/rules.d/66-snapd-autoimport.rules
@@ -1,3 +1,2 @@
-# probe for assertions, must run before udisks2
-ACTION=="add", SUBSYSTEM=="block", KERNEL!="loop*", KERNEL!="ram*" \
-    RUN+="/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/%k"
+ACTION=="add|change", SUBSYSTEM=="block", KERNEL!="loop*", KERNEL!="ram*" \
+    ENV{SYSTEMD_WANTS}+="snapd.autoimport-device@%k.service"


valentindavid:
To test it using qemu, it is possible to simulate usb insertion of usb disk. Just add a drive in command line with `if=none` and an id. Then in the qemu console, use `device_add usb-storage,drive=<id>`

static/usr/lib/systemd/system/snapd.autoimport-device@.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Auto import assertions from a specific block device
+After=snapd.service snapd.socket snapd.seeded.service
+# don't run on classic
+ConditionKernelCommandLine=|snap_core
```
alfonsosanchezbeato, line 5: We don't have `snap_core` in the kernel command line in UC20+ iirc. I see:
`snapd_recovery_mode=run console=ttyS0 console=tty1 panic=-1`
I don't think we need to do this backward compatible.

static/usr/lib/systemd/system/snapd.autoimport-device@.service
```diff
@@ -0,0 +1,12 @@
...
+# TODO:UC20: only enable this in run mode?
+ConditionKernelCommandLine=|snapd_recovery_mode
```
alfonsosanchezbeato, line 7: @anonymouse64 wdyt about this?

alfonsosanchezbeato (commented):
Thanks for this! I have a couple of comments. Also, we need to think about the way somebody would set `SNAP_AUTO_IMPORT` at a high level. I guess it would be possible to do it by introducing a rules files and using the `system-files` assertion, but it would be probably better to do it via `snap set system ...`.

Another question, what happens if you introduce/extract the pen drive multiple times? Have you tried that?

Finally, are there snapd tests for auto-import? If they are, we could add those to the list of snapd tests that we run.

static/usr/lib/systemd/system/snapd.autoimport-device@.service
valentindavid: Right. I copied that from from `snapd.autoimport.service`. I will fix that.

static/usr/lib/systemd/system/snapd.autoimport-device@.service
valentindavid: I will set it to `snapd_recovery_mode=run`. Recovery and install should not do anything interesting. Unless someone wants to login during the install.

static/usr/lib/systemd/system/snapd.autoimport-device@.service
valentindavid: Removed

static/usr/lib/systemd/system/snapd.autoimport-device@.service
valentindavid: Changed

alfonsosanchezbeato: Right, I'm +1 on this. I think that you can get a shell in recovery without this iirc?

static/usr/lib/udev/rules.d/66-snapd-autoimport.rules
```diff
@@ -1,3 +1,2 @@
-# probe for assertions, must run before udisks2
-ACTION=="add", SUBSYSTEM=="block", KERNEL!="loop*", KERNEL!="ram*" \
-    RUN+="/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/%k"
+ACTION=="add|change", SUBSYSTEM=="block", KERNEL!="loop*", KERNEL!="ram*", ENV{SNAP_AUTO_IMPORT}!="0" \
+    ENV{SYSTEMD_WANTS}+="snapd.autoimport-device@%k.service"
```
anonymouse64, line 2: so to be clear `SYSTEMD_WANTS` here will trigger execution of a service named like `snapd.autoimport-device@foo.service` which then results in the ExecStart executing `snap auto-import --mount=/dev/foo` correct?

static/usr/lib/udev/rules.d/66-snapd-autoimport.rules
```diff
@@ -1,3 +1,2 @@
-# probe for assertions, must run before udisks2
-ACTION=="add", SUBSYSTEM=="block", KERNEL!="loop*", KERNEL!="ram*" \
-    RUN+="/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/%k"
+ACTION=="add|change", SUBSYSTEM=="block", KERNEL!="loop*", KERNEL!="ram*", ENV{SNAP_AUTO_IMPORT}!="0" \
```
anonymouse64, line 1: I have the same question as Alfonso what will set SNAP_AUTO_IMPORT on the device?

Note that in general, we do not permit arbitrary udev rules to be installed on ubuntu core, there is a design effort underway at some point to allow certain kinds of udev rules to be installed via the gadget snap, but that is still in its infancy phase (this is on @pedronis plate, not sure how far along the design/thinking is on that). Due to this, I would be inclined to not introduce this functionality until we have a use case for it and we know how things will fit together.

anonymouse64 (commented):
In general, I'm +1 on UC22 using systemd for importing devices, but I don't think we need the SNAP_AUTO_IMPORT functionality

static/usr/lib/udev/rules.d/66-snapd-autoimport.rules
valentindavid: Yes.

valentindavid: But it is scheduled instead. udevd does not wait for it.

static/usr/lib/udev/rules.d/66-snapd-autoimport.rules
valentindavid: For instance if a user wants to use udisks2, they will provide it in a snap. The problem is that there will be error with udisks2 daemon and `snap auto-import` conflicting on mounting the filesystem. They can provide an udev rule with `UDISKS_AUTO=0` to disable auto mounting of some filesystems by udisks2. But there is nothing to disable `snap auto-import`. So that means there is no way to make some device being used only by udisks2. We need a way for them to disable `snap auto-import`. They can do it the way they want. For example they could do, only use `snap auto-import` if the fs label is "ASSERT":

```
# Only added/inserted block devices
ACTION=="add|change", SUBSYSTEM=="block", GOTO="start"
GOTO="end"
LABEL="start"
# Import blkid, or this is not useful if run after 60-persistent-storage.rules
IMPORT{builtin}="blkid"
# Check if the filesystem label is "ASSERT"
ENV{ID_FS_LABEL}=="ASSERT", GOTO="found_assert"
# If not, disable snap auto-import
ENV{SNAP_AUTO_IMPORT}=0
GOTO="end"
# If it is, disable udisks2 auto mount (can also use UDISKS_IGNORE)
LABEL="found_assert"
ENV{UDISKS_AUTO}="0"
LABEL="end"
```

anonymouse64: I guess what I'm trying to say is this step:

> They can provide an udev rule with UDISKS_AUTO=0 to disable auto mounting of some filesystems by udisks2

Is only doable if you have an unconfined shell on an Ubuntu Core device which isn't something we support, we really mainly focus on people modifying the system via confined snaps which only can install udev rules as per the interfaces that their snaps connect.

alfonsosanchezbeato:
To be clear, the only problem I see with the PR at the moment is the introduction of the `SNAP_AUTO_IMPORT` env variable before having a design on how we will make auto-import compatible with udisks2. I would prefer to remove that bit for the moment and assume that udisks2 cannot be used yet with UC22, and probably open a bug/issue so we can track the problem. We need auto-import to be working ASAP (we use it in CI/CD), and we can wait a bit on the solution for udisks2.

valentindavid:
I have removed `SNAP_AUTO_IMPORT`.

alfonsosanchezbeato (approved):
LGTM, thanks

Meulengracht (approved):
This looks fine to me as well, but as @alfonsosanchezbeato asked, it would be nice to know if there are any tests in the snapd suite that verifies the auto-import, as it seems like something that would be testable - @mvo5 do you know?

valentindavid:
Yes, there are tests in snapd. This is how we discovered the bug. They are disabled for core 22, in snapcore/snapd#11476 for now. But when this is PR is merged and released, we can finally enable them.

valentindavid:
See `tests/core/snap-auto-mount/task.yaml`

bboozzoo:
Does this PR need a backport to core20 too?

valentindavid:
> Does this PR need a backport to core20 too?

I do not think udevd in systemd version in core20 blocks mounts in programs launched from udevd. So it is not required.

PR #36, opened 2022-03-16
probert: from git instead of archive

dbungert:
LP: #1964571

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -14,7 +14,7 @@ parts:
     plugin: nil
     source: https://github.com/canonical/probert.git
     source-type: git
-    source-branch: server/jammy
+    source-branch: 639255e5ddceb00cb21c2436e7ce1f7430aa710f
     override-pull: |
       snapcraftctl pull
       # install build dependencies
@@ -48,9 +48,9 @@ parts:
     after:
       - probert-deb
     plugin: nil
-    source: https://github.com/CanonicalLtd/subiquity.git
+    source: https://github.com/canonical/subiquity.git
     source-type: git
-    source-branch: main
+    source-branch: 8f9f97c69bb36048ea4c9e16be34bca3ac141771
     override-pull: |
       snapcraftctl pull
       # install build dependencies
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -14,7 +14,7 @@ parts:
     plugin: nil
     source: https://github.com/canonical/probert.git
     source-type: git
-    source-branch: 639255e5ddceb00cb21c2436e7ce1f7430aa710f
+    source-commit: 639255e5ddceb00cb21c2436e7ce1f7430aa710f
     override-pull: |
       snapcraftctl pull
       # install build dependencies
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -50,7 +50,7 @@ parts:
     plugin: nil
     source: https://github.com/canonical/subiquity.git
     source-type: git
-    source-branch: 8f9f97c69bb36048ea4c9e16be34bca3ac141771
+    source-commit: 8f9f97c69bb36048ea4c9e16be34bca3ac141771
     override-pull: |
       snapcraftctl pull
       # install build dependencies
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -10,7 +10,41 @@ build-base: core22
 grade: devel
 
 parts:
+  probert-deb:
+    plugin: nil
+    source: https://github.com/canonical/probert.git
+    source-type: git
+    source-branch: server/jammy
+    override-pull: |
+      snapcraftctl pull
+      # install build dependencies
+      export DEBIAN_FRONTEND=noninteractive
+      export DEBCONF_NONINTERACTIVE_SEEN=true
+      sudo -E apt-get build-dep -y ./
+    override-build: |
+      # unset the LD_FLAGS and LD_LIBRARY_PATH vars that snapcraft sets for us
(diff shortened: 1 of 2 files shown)


dbungert:
@alfonsosanchezbeato FYI

snapcraft.yaml
```diff
@@ -10,7 +10,41 @@ build-base: core22
...
+      # TODO: should we unset $PATH to not include $SNAPCRAFT_STAGE too?
+      unset LD_FLAGS
+      unset LD_LIBRARY_PATH
+      # run the real build (but just build the binary package, and don't
+      # bother compressing it too much)
+      dpkg-buildpackage -b -uc -us -Zgzip -zfast
+      cp ../probert-{network,common}*.deb $SNAPCRAFT_PART_INSTALL
+    stage:
+      - -probert*.deb
+    override-prime: |
```
alfonsosanchezbeato, line 38: I think it is appropriate to add the same comment as in the consoleconf-deb part:
```suggestion
    # XXX: Dirty hacks to enable building core22 on non-focal systems.
    # See below for details.
    override-prime: |
```

alfonsosanchezbeato (commented):
Thanks for this! Waiting for the tests to finish to approve (I have also a minor suggestion). However, I am a bit worried now about how we are pulling source for subiquity and now for probert too. Ideally we would like to get things from tags instead of branches to have more control on what is pulled in the base. Do you have any suggestion on which tags we should use for that?

alfonsosanchezbeato (approved):

alfonsosanchezbeato:
I have locally tested this on my NUC with wifi, successfully.

dbungert:
Updated to use commit hashes instead of branch references.
I'm not opposed to tags, that would also be fine.

alfonsosanchezbeato:
Thanks for the changes - yeah, commits are good for me as well. Maybe depending on the policy you follow for creating releases/tags for subiquity/probert we can use release tags in the future.

snapcraft.yaml
```diff
@@ -10,11 +10,47 @@ build-base: core22
 grade: devel
 
 parts:
+  probert-deb:
+    plugin: nil
+    source: https://github.com/canonical/probert.git
+    source-type: git
+    source-branch: 639255e5ddceb00cb21c2436e7ce1f7430aa710f
```
alfonsosanchezbeato, line 17:
```suggestion
    source-commit: 639255e5ddceb00cb21c2436e7ce1f7430aa710f
```

snapcraft.yaml
```diff
@@ -10,11 +10,47 @@ build-base: core22
...
+      snapcraftctl prime
   consoleconf-deb:
+    after:
+      - probert-deb
     plugin: nil
-    source: https://github.com/CanonicalLtd/subiquity.git
+    source: https://github.com/canonical/subiquity.git
     source-type: git
-    source-branch: main
+    source-branch: 8f9f97c69bb36048ea4c9e16be34bca3ac141771
```
alfonsosanchezbeato, line 53:
```suggestion
    source-commit: 8f9f97c69bb36048ea4c9e16be34bca3ac141771
```

snapcraft.yaml
dbungert: Thanks for fixing that up!

valentindavid (commented):
It would be nice to comment next to the "source-commit" entries that the branch we are using is `server/jammy`. Also it would be nice to have a small explanation of why, or when we can change the branch to `main` again.

dbungert:
@valentindavid - Comments added.

Tracking `main` of Subiquity is not something I would recommend for a stable release core snap - Subiquity is under pretty active development at the moment for a variety of topics, so I think the odds of undesired regressions for console-conf are higher than the odds of beneficial fixes.  

Server team requested that we implement a stable branch for Subiquity and follow the distribution feature freeze timelines, which we have implemented.  So if one wanted to follow a branch instead, that stable branch (`server/jammy` at the moment), would be my recommendation.

When work is eventually started for core24, perhaps that would be a good moment to point to the `main` branch.

What do you think?

alfonsosanchezbeato:
> @valentindavid - Comments added.
> 
> Tracking `main` of Subiquity is not something I would recommend for a stable release core snap - Subiquity is under pretty active development at the moment for a variety of topics, so I think the odds of undesired regressions for console-conf are higher than the odds of beneficial fixes.
> 
> Server team requested that we implement a stable branch for Subiquity and follow the distribution feature freeze timelines, which we have implemented. So if one wanted to follow a branch instead, that stable branch (`server/jammy` at the moment), would be my recommendation.
> 
> When work is eventually started for core24, perhaps that would be a good moment to point to the `main` branch.
> 
> What do you think?

It seems sensible to me - the only question I have is, when do you tag the branches? Ideally we will want to follot tags instead of commits, when available.

dbungert:
> It seems sensible to me - the only question I have is, when do you tag the branches? Ideally we will want to follow tags instead of commits, when available.

Today we do tags as part of a Subiquity stable release, which is a process that is somewhat disconnected from the distribution release timelines.  Subiquity has a refresh capability to see the newer snap on the store and to update the live environment to that newer version, so we try to take advantage of that every few months.

Is the intention of the tag something similar to my usage of a direct commit hash, albeit one that would be more descriptive?  Or do you have something else in mind?

alfonsosanchezbeato:
> Is the intention of the tag something similar to my usage of a direct commit hash, albeit one that would be more descriptive? Or do you have something else in mind?

Correct, basically the same purpose, but more descriptive so it is easier to know, for instance, how far are we from latest release, etc.

valentindavid (approved):

PR #40, opened 2022-04-11
core: add pkcs11 dependencies

kubiko:
Add pkcs11 dependencies into core20 base snap

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -68,17 +68,24 @@ PACKAGES=(
     dmsetup
     dosfstools
     finalrd
+    gnutls-bin
     iproute2
     iptables
     iputils-ping
     kmod
     less
+    libengine-pkcs11-openssl
     libnss-extrausers
+    libopts25
+    libp11-3
     libpam-modules
     libpam-systemd
     netcat-openbsd
     netplan.io
+    opensc
     openssh-server
+    p11-kit
+    p11-kit-modules
     rfkill
     squashfs-tools
     sudo
--- a/hooks/400-trim-pkcs-11.chroot
+++ b/hooks/400-trim-pkcs-11.chroot
@@ -1,0 +1,81 @@
+#!/bin/sh -ex
+
+mkdir -p /usr/share/p11-kit/modules
+echo "Creating p11-kit-client module configuration"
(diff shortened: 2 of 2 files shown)


xnox:
As far as i understand this will add pkcs#11 dependencies/libraries to the core22 snap; and it will also enable p11-kit-client provider in the default proxy pkcs#11 provider, which will try to connect to local socket which p11-kit-server should provide. I am assuming that sockets will be provided by something external/device specific.

module: opensc-pkcs11.so is available too, which one can opt into by configuring individual user account (via ~/.config), or system (via /etc) to use.

I'm not sure if it makes sense to have p11-kit-client module enabled by default in core, or if custom devices that provide p11-kit-server can also configure and activate p11-kit-client module.

Hopefully I understand all of the above correctly.

valentindavid:
I am surprised by `/usr/lib/x86_64-linux-gnu/engines-3/pkcs11.la`. I do not think that openssl uses `lt_dlopen`, so it must be a useless file. Is that an oversight in `libengine-pkcs11-openssl` Ubuntu package?

Not sure about the use for `/etc/xdg/autostart/pkcs11-register.desktop`.

What is libopts for?

There are 2 dead symlinks in `/usr/lib/x86_64-linux-gnu/pkcs11/`.

Also there are 2 bash completion configuration that might not be useful: `goid-tool` and `pkcs11-register`

valentindavid:
Did not we want `opensc-pkcs11.so` to be only used in snaps that manage the smart cards/dongles. It feels that it should not be in core. Maybe we should remove all `opensc-pkcs11` (it is a dependency of opensc)

kubiko:
Rebase and push to remove dependency on ppa.

The agreement is to go ahead and use the pkcs11-tool from the archive as it is, to maintain current behaviour.
Most of the use cases are using openssl which uses libp11 which already defaults to the p11-proxy.

kubiko:
@valentindavid fair point about `opensc-pkcs11.so` and it'd also reming anybody calling `pkcs11-tool` that default module is not there.

valentindavid (approved):
Same comments as core20 PR.

xnox (approved):

xnox:
Note that i have no powers to merge, build, or release this change.

PR #41, opened 2022-04-20
hooks: Add symlinks for bash completions of snap and snaps

valentindavid:
Completion for `snap` must be present in
`/usr/share/bash-completion/completions/` and it must point to the
file in the active Snapd.

Other completion will be added in
`/var/lib/snapd/desktop/bash-completion/completions/` by Snapd. To be
able to access those `XDG_DATA_DIRS` has to be set by Snapd's
`profile.d` file.

This commits also adds snap's `environment.d` file which is used
by user services rather than interactive users.

This depends on snapcore/snapd#11038

### DIFF
--- a/hooks/050-snap-symlink.chroot
+++ b/hooks/050-snap-symlink.chroot
@@ -2,3 +2,12 @@
 
 echo "Creating the snap binary symlink"
 ln -s /snap/snapd/current/usr/bin/snap /usr/bin/snap
+
+echo "Creating the symlink to snap's bash completion"
+ln -s /snap/snapd/current/usr/share/bash-completion/completions/snap /usr/share/bash-completion/completions/snap
+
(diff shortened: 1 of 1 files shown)


xnox:
@valentindavid the snapd pull request is merged, so we should land this too?

alfonsosanchezbeato (approved):
LGTM - I understand this can land even if the snapd changes are not in stable yet, right? As you would simply get some dangling links that won't be used.

valentindavid:
> LGTM - I understand this can land even if the snapd changes are not in stable yet, right? As you would simply get some dangling links that won't be used.

Those files already exist in snapd. So no dangling links. No error message. Just that the completion files do not get installed and are not found.

valentindavid:
The needed feature was released in snapd 2.57. We can merge.

PR #44, opened 2022-05-13
Fix image download URL

valentindavid:


Meulengracht (approved):
LGTM

Makefile
```diff
@@ -1,11 +1,7 @@
...
-#LTS=jj
-#BASE := $(LTS)-base-$(DPKG_ARCH).tar.gz
-#URL := http://cdimage.ubuntu.com/ubuntu-base/$(LTS)/daily/current/$(BASE)
-# Temporarily until JJ is released
-DEVEL=jammy
-BASE := $(DEVEL)-base-$(DPKG_ARCH).tar.gz
-URL := http://cdimage.ubuntu.com/ubuntu-base/daily/current/$(BASE)
+LTS=22.04
+BASE := ubuntu-base-$(LTS)-base-$(DPKG_ARCH).tar.gz
+URL := http://cdimage.ubuntu.com/ubuntu-base/releases/$(LTS)/release/$(BASE)
```
alfonsosanchezbeato, line 4: In the past, we downloaded the base from `http://cdimage.ubuntu.com/ubuntu-base/$(LTS)/daily/current/$(BASE)`, and I see that in  `http://cdimage.ubuntu.com/ubuntu-base/` there are `bionic` and `focal` subdirectories, but no `jammy` one. But of course we need to change this as now there is no jammy base in the current URL, only kinetic. The problem is that the base in the `releases` folder is not updated, and although we do update/upgrade in the hooks, it is not ideal. @sil2100 will we have a `http://cdimage.ubuntu.com/ubuntu-base/jammy/daily/curren/` folder with daily bases in the near future?

valentindavid: @alfonsosanchezbeato The repository does not build without this change and we cannot merge anything until this is fixed. Could we merge this and keep this discussion as an issue to be resolved later?

alfonsosanchezbeato: Ok, I've opened https://github.com/snapcore/core-base/issues/46 to track this.

alfonsosanchezbeato (approved):
👍

PR #48, opened 2022-05-24
Use lzo compression

valentindavid:
This seems to reduce the boot on rpi3 by 10s.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -9,6 +9,7 @@ type: base
 build-base: core22
 grade: stable
 assumes: [snapd2.55.5]
+compression: lzo
 
 parts:
   probert-deb:


valentindavid:
> Also, could you please comment on size increment with the change?

On x86_64, it goes from 71M to 93M.

mvo5:
This feels like a risky change to me, at least right now. We don't have enough telemetry currently to know if all core devices can take the size increase. And while snapd has some low-diskspace checks I'm not sure they are good enough. It's also something to double with David Beamonte because (relatively) small size is a selling point so we need to consider this trade-off carefully.

mvo5:
Thank you, I think this is something we need to re-evaulate for uc24 but right now it feels like too early.

PR #49, opened 2022-06-03
static/usr/lib/core: mount ubuntu-save to /var/lib/snapd/save

Meulengracht:
As a part of the customer request to have a per-snap folder on the ubuntu-save partition, and because we do not want to expose the /run/mnt path to apparmor, we instead mount ubuntu-save into /var/lib/snapd as 'save'. This is already done by snap, but this happens to late in the startup process, and would create a dependency between snaps and snapd.

### DIFF
--- a/static/usr/lib/core/handle-writable-paths
+++ b/static/usr/lib/core/handle-writable-paths
@@ -181,6 +181,7 @@ main()
 		echo "/dev/root / rootfs defaults,ro 0 0" >> "$fstab"
 		handle_writable_paths "$writable_paths" "$fstab"
 		echo "/run/mnt/ubuntu-seed /var/lib/snapd/seed none bind,ro 0 0" >> "$fstab"
+		echo "/run/mnt/ubuntu-save /var/lib/snapd/save none bind,rw 0 0" >> "$fstab"
 	fi
 
 	# IMPORTANT: ensure we synced everything back to disk
--- a/static/usr/lib/core/handle-writable-paths
+++ b/static/usr/lib/core/handle-writable-paths
@@ -181,7 +181,6 @@ main()
 		echo "/dev/root / rootfs defaults,ro 0 0" >> "$fstab"
 		handle_writable_paths "$writable_paths" "$fstab"
 		echo "/run/mnt/ubuntu-seed /var/lib/snapd/seed none bind,ro 0 0" >> "$fstab"
-		echo "/run/mnt/ubuntu-save /var/lib/snapd/save none bind,rw 0 0" >> "$fstab"
 	fi
 
 	# IMPORTANT: ensure we synced everything back to disk
--- a/static/usr/lib/systemd/system/var-lib-snapd-save.mount
+++ b/static/usr/lib/systemd/system/var-lib-snapd-save.mount
@@ -1,0 +1,8 @@
+[Unit]
+ConditionPathIsMountPoint=/run/mnt/ubuntu-save
+
+[Mount]
+What=/run/mnt/ubuntu-save
+Where=/var/lib/snapd/save
+Options=bind
+Type=none
--- a/tests/spread/main/basic/task.yaml
+++ b/tests/spread/main/basic/task.yaml
@@ -43,5 +43,7 @@ execute: |
   execute_remote systemctl -q is-active systemd-timesyncd.service
 
   # Verify that ubuntu-seed and ubuntu-save has been mounted
-  execute_remote mount | awk '{if ($3 == "/var/lib/snapd/seed") { exit 0}} ENDFILE{exit -1}'
-  execute_remote mount | awk '{if ($3 == "/var/lib/snapd/save") { exit 0}} ENDFILE{exit -1}'
+  # ubuntu-seed should be mounted readonly
+  # ubuntu-save should be mounted readwrite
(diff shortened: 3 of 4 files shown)


tests/spread/main/basic/task.yaml
```diff
@@ -41,3 +41,7 @@ execute: |
   execute_remote systemctl -q is-active systemd-journald.service
   execute_remote systemctl -q is-active systemd-resolved.service
   execute_remote systemctl -q is-active systemd-timesyncd.service
+
+  # Verify that ubuntu-seed and ubuntu-save has been mounted
+  execute_remote mount | awk '{if ($3 == "/var/lib/snapd/seed") { exit 0}} ENDFILE{exit -1}'
+  execute_remote mount | awk '{if ($3 == "/var/lib/snapd/save") { exit 0}} ENDFILE{exit -1}'
```
valentindavid, line 47: Should we check it is mounted read/write? If snapd mounts that path in read-only, I would expect this line to not fail.

Meulengracht: Added, thanks for this suggestion!

valentindavid (commented):
Maybe the test should compare `systemctl show -p ActiveEnterTimestampMonotonic --value` for both `local-fs.target` and `var-lib-snapd-save.mount`. `var-lib-snapd-save.mount` should have been active before `local-fs.target`.

static/usr/lib/core/handle-writable-paths
```diff
@@ -181,6 +181,7 @@ main()
 		echo "/dev/root / rootfs defaults,ro 0 0" >> "$fstab"
 		handle_writable_paths "$writable_paths" "$fstab"
 		echo "/run/mnt/ubuntu-seed /var/lib/snapd/seed none bind,ro 0 0" >> "$fstab"
+		echo "/run/mnt/ubuntu-save /var/lib/snapd/save none bind,rw 0 0" >> "$fstab"
```
pedronis, line 184: ubuntu-save is optional if encryption is not enabled

we also need the same change in core20?

valentindavid: > ubuntu-save is optional if encryption is not enabled

I suppose it is always created. But if it is removed, it should still work. Then maybe the tests should try removing the partition and see if still boots.

Without `nofail`, it should in theory fail to boot.

Meulengracht: > 

Interesting point about `nofail`, I believe it would make sense to put that

pedronis: Well that will continue the boot even if the device is there but the mount itself fail. My preference is to add that line only if the /run/mnt/ubuntu-save exists

static/usr/lib/systemd/system/var-lib-snapd-save.mount
```diff
@@ -0,0 +1,8 @@
+[Unit]
+ConditionPathIsMountPoint=/run/mnt/ubuntu-save
+
+[Mount]
+What=/run/mnt/ubuntu-save
+Where=/var/lib/snapd/save
```
pedronis, line 6: does this doesn't need any requirement keywords because systemd does the right thing in terms of dependencies checking having var/lib mounted before trying this?

valentindavid: This is the effective properties of the unit (only dependencies):

```
Requires="dev-mapper-ubuntu\\x2dsave\\x2dbafd47f3\\x2da887\\x2d4061\\x2d997f\\x2dcf8a073b38bb.device" "run-mnt-ubuntu\\x2dsave.mount" system.slice var-lib-snapd.mount -.mount
RequiredBy=local-fs.target
Conflicts=umount.target
Before=umount.target local-fs.target
After="blockdev@dev-mapper-ubuntu\\x2dsave\\x2dbafd47f3\\x2da887\\x2d4061\\x2d997f\\x2dcf8a073b38bb.target" "run-mnt-ubuntu\\x2dsave.mount" "dev-mapper-ubuntu\\x2dsave\\x2dbafd47f3\\x2da887\\x2d4061\\x2d997f\\x2dcf8a073b38bb.device" systemd-journald.socket system.slice var-lib-snapd.mount -.mount local-fs-pre.target
StopPropagatedFrom="dev-mapper-ubuntu\\x2dsave\\x2dbafd47f3\\x2da887\\x2d4061\\x2d997f\\x2dcf8a073b38bb.device"
RequiresMountsFor=/run/mnt/ubuntu-save /var/lib/snapd
```

So I am not sure where it is documented that `RequiresMountsFor` is automatic.

Meulengracht: *If a mount unit is beneath another mount unit in the file system hierarchy, both a requirement dependency and an ordering dependency between both units are created automatically.* AFAIU systemd will automatically detect any filesystem tree dependencies and order/depend on those

valentindavid: It is state in the "implicit dependencies" of the `systemd.mount` manual page.

pedronis (approved):
thank you

alfonsosanchezbeato (approved):
LGTM, thanks

PR #51, opened 2022-06-13
console-conf: update commit for LP: #1978339

dbungert:
Not moving this to a tag yet, as this fix is console-conf specific.
Picks up https://github.com/canonical/subiquity/pull/1310

PR #52, opened 2022-06-15
subiquity: update to 22.05.1

valentindavid:
Also remove hook script that was working around an old version of subiquity.


valentindavid:
I did not see #51. I close, will re-open the fix separately.

PR #54, opened 2022-06-15
hooks: remove hook working around missing parameter to agetty

valentindavid:
console-conf upstream service already has `-i`, so now we were adding
it a second time.


alfonsosanchezbeato (approved):
LGTM

PR #56, opened 2022-06-15
Verify and cache base image download

valentindavid:
Downloading though plain HTTP without verifying signed hash might make
it potentially vulnerable if our builders get attacked with for
example route hijacking or DNS poisoning.

Also making separate snapcraft part makes snapcraft not to re-download
every time we need to rebuild locally. Just use `snapcraft clean base`
to force re-downloading before.


alfonsosanchezbeato (approved):
LGTM, thanks

PR #57, opened 2022-06-15
snapcraft.yaml: Fix part dependencies

valentindavid:
`consoleconf-deb` does not need to rebuild if `probert-deb` was
changed.


Meulengracht (approved):
Nice, thanks!

PR #58, opened 2022-06-15
many: add configuration for splash screen

alfonsosanchezbeato:
Adds configuration to the base for splash screen. This requires also initramfs changes. Just a draft for the moment as we need a new plymouth deb package in jammy and some snapd changes.

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -69,6 +69,7 @@ PACKAGES=(
     dosfstools
     finalrd
     gnutls-bin
+    fonts-ubuntu
     iproute2
     iptables
     iputils-ping
@@ -86,6 +87,7 @@ PACKAGES=(
     openssh-server
     p11-kit
     p11-kit-modules
+    plymouth-label-ft
     rfkill
     squashfs-tools
     sudo
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -19,25 +19,45 @@ trap 'umount /proc' EXIT
 # systemd postinst needs this
 mkdir -p /var/log/journal
 
+# We need to install first the certificates as apt will need them for https to lp
+apt update
+apt install --no-install-recommends -y ca-certificates
+
 # shellcheck disable=SC1091
 CODENAME=$(. /etc/os-release; echo "$UBUNTU_CODENAME")
-# enable the foundations ubuntu-image PPA
-echo "deb http://ppa.launchpadcontent.net/canonical-foundations/ubuntu-image/ubuntu $CODENAME main" > /etc/apt/sources.list.d/ubuntu-image.list
+# enable the ucdev PPA with additional packages to build bases
+echo "deb https://ppa.launchpadcontent.net/ucdev/base-ppa/ubuntu/ $CODENAME main" \
+     > /etc/apt/sources.list.d/ubuntu-image.list
 
-cat >/etc/apt/trusted.gpg.d/canonical-foundations-ubuntu-image.asc <<EOF
+cat >/etc/apt/trusted.gpg.d/ucdev-base-ppa.asc <<'EOF'
 -----BEGIN PGP PUBLIC KEY BLOCK-----
 
-xo0EUL4ncAEEAOZssKpJDMZKbmsf9lHwlKA0vN6yQ0sOIPc500waH3xTC0sVlqQc
-3pUxCIdhU+qK1mH2D51FGHDb504k0Lpb+LE56TWa/X3xrZqUQX0UD1fykEruR4W2
-CdkXXZvmNBNatE9GurR6p407X5TED+dlUK/hIKNCb5unTEilBb4WwArxABEBAAHN
-LExhdW5jaHBhZCBQUEEgZm9yIENhbm9uaWNhbCBGb3VuZGF0aW9ucyBUZWFtwrgE
-EwECACIFAlC+J3ACGwMGCwkIBwMCBhUIAgkKCwQWAgMBAh4BAheAAAoJENTAtmj9
-TJE5u/MD/2j2auOv62YUFwT7POylj7ihhZOarOSCEiQGita8II77j5AoK5O75uD+
-oQc5pdxVN2NGYD5R0PmDCPFN1Rb869YjtsPgLefEB+6Tc1GOR9hgnwuSU5lrwqdQ
-Ht/skh2wZSHtJgejt9kqIKMho1wtYz7ZTqMtN9GJK0VONbHP0Xu6
-=lWSn
+mQINBGLW9dwBEAC7UJatAWOrUgfJlElkTUfmjL6JCTONYGy1Qzrw65YgzjdJGvXQ
+iuOZyeX3lGjBrihbH7hQvlf2Cyo6E7DA/PLUMVAUMfc52U9RH2ovejZ25BonyL/e
+XArTBRNWcjOgZXsO2aAcQZ7Tn1hYbjYv8zGmurk1vbQYKd/LlZsdrAkq2lmVOFPq
+MTDLfa8vIhlfbr7uQlcB6uBehLYUUoTW3nV9HiA6fDTr86N2cQjhP7zxTZHxn52p
+/ba0um5Hlg814l80S3ALaPQzqnpPPix0yzT7P2JfoIXKof4Lm28I2txK3fATeFzI
+AQpqZqsCyRxyNJcN3AW9ma/dz7Orgq0y0q0xzTtvHq3+MuVAwFRhk9wyOZNPU9a6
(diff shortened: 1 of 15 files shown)


alfonsosanchezbeato:
This is ready for an initial pass now. Note that to have the full feature the changes in the initramfs are needed, and a gadget with additional `quiet splash` kernel command line parameters is required.

hooks/026-configure-plymouth.chroot
```diff
@@ -0,0 +1,141 @@
+#! /bin/sh -ex
+
+# Rewrite units for plymouth
+cat << 'EOF' > /lib/systemd/system/plymouth-start.service
```
valentindavid, line 4: I feel like we should use drop-ins to tweak upstream units, that would be easier to maintain.

valentindavid:
`plymouth-{halt,poweroff,reboot,kexec}.service` are missing `--ignore-serial-consoles`. Without it, it seems to not display anything when shutting down.

hooks/026-configure-plymouth.chroot
alfonsosanchezbeato: Changed now to use overrides

alfonsosanchezbeato:
> `plymouth-{halt,poweroff,reboot,kexec}.service` are missing `--ignore-serial-consoles`. Without it, it seems to not display anything when shutting down.

Added `--ignore-serial-consoles` in the places where it was not yet.

valentindavid:
Do we need `/usr/libexec/plymouth/plymouth-generate-initrd`, `/usr/libexec/plymouth/plymouth-populate-initrd`, `/usr/libexec/plymouth/plymouth-update-initrd`?

Also I have noticed we have files in `/usr/share/initramfs-tools`.

static/usr/lib/systemd/system/plymouth-start.service.d/core-override.conf
```diff
@@ -0,0 +1,9 @@
+[Unit]
+After=snap-initramfs-mounts.service
+Requires=plymouth-quit.service splash-client.service
+Requires=plymouth-show-run.service plymouth-show-install.service
+Requires=plymouth-show-recover.service plymouth-show-factory-reset.service
```
valentindavid, line 5: I would prefer `Wants`. Specially if we move `plymouth-quit.service` from `multi-user.target.wants`.

hooks/026-configure-plymouth.chroot
```diff
@@ -0,0 +1,109 @@
...
+
+# Note that we have overrides for plymouth services in:
+# static/usr/lib/systemd/system/plymouth-start.service.d/core-override.conf
+# static/usr/lib/systemd/system/plymouth-quit.service.d/core-override.conf
+# static/usr/lib/systemd/system/plymouth-halt.service.d/core-override.conf
+# static/usr/lib/systemd/system/plymouth-reboot.service.d/core-override.conf
+# static/usr/lib/systemd/system/plymouth-kexec.service.d/core-override.conf
+# static/usr/lib/systemd/system/plymouth-poweroff.service.d/core-override.conf
+
+cat << 'EOF' > /lib/systemd/system/splash-client.service
```
valentindavid, line 11: Those files are not dynamic. Could we move them into `static/`?

valentindavid: Maybe we can keep the configuration files here. But the systemd units would be better in `static/`.

hooks/026-configure-plymouth.chroot
```diff
@@ -0,0 +1,109 @@
...
+ConditionKernelCommandLine=snapd_recovery_mode=run
+ConditionPathExists=/run/plymouth/pid
+
+[Service]
+ExecStart=-/bin/plymouth display-message --text='Starting system...'
+Type=oneshot
+TimeoutSec=20
+
+[Install]
+WantedBy=sysinit.target
```
valentindavid, line 51: I am not sure why there is an `[Install]` section if it there is `Requires` added in `plymouth-start.service`

static/usr/bin/splash-client
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -exu
```
valentindavid, line 1: Maybe the `x` can be removed.

valentindavid: If you were using for debugging, I suppose.

static/usr/lib/systemd/system/plymouth-quit.service.d/core-override.conf
```diff
@@ -0,0 +1,6 @@
+[Unit]
+After=snapd.seeded.service
+ConditionKernelCommandLine=snapd_recovery_mode=run
+
+[Service]
+ExecStartPost=-systemctl stop splash-client.service
```
valentindavid, line 6: I would use `plymouth --ping` in the loop of `splash-client` instead so we do not have to change `plymouth-quit.service`.

static/usr/lib/systemd/system/plymouth-start.service.d/core-override.conf
```diff
@@ -0,0 +1,9 @@
...
+
+[Service]
+ExecStart=
+ExecStart=/usr/sbin/plymouthd --debug --mode=boot --pid-file=/run/plymouth/pid --attach-to-session --ignore-serial-consoles
```
valentindavid, line 9: I suppose we can remove `--debug`.

static/usr/bin/splash-client
```diff
@@ -0,0 +1,15 @@
...
+
+# Change 1 is "Initialize system state"
+msg=
+prev_msg=
+while true; do
+    sleep 0.5
+    msg=$(snap change --abs-time 1 | grep -m1 Doing |
```
valentindavid, line 8: Is this suppose to show messages only on the first boot?

valentindavid: Would `snap tasks --last=seed? --abs-time` be more obvious?

Should we also check `--last=become-operational`?

static/usr/bin/splash-client
```diff
@@ -0,0 +1,15 @@
...
+              awk -F ' +' '{ s = ""; for (i = 4; i <= NF; i++) s = s $i " "; print s }') \
```
valentindavid, line 9: This is fine. But here are some alternatives:

Without awk:
```
tr -s " " | cut -d" " -f4-
```

Merging awk and grep together, and removing the loop (which also does not squash multiple spaces):
```
awk -F " +" '$1 == "Doing" { print substr($0, index($0,$4)); exit }'
```

valentindavid: And for the lol in sed:
```
sed -E '/^Doing /{;s/([^ ]*[ ]*){3}//;q};d'
```

hooks/001-extra-packages.chroot
```diff
@@ -19,25 +19,45 @@ trap 'umount /proc' EXIT
...
+# We need to install first the certificates so the cert for the PPA is trusted
+apt update
+apt install --no-install-recommends -y ca-certificates
+
 # shellcheck disable=SC1091
 CODENAME=$(. /etc/os-release; echo "$UBUNTU_CODENAME")
-# enable the foundations ubuntu-image PPA
-echo "deb http://ppa.launchpadcontent.net/canonical-foundations/ubuntu-image/ubuntu $CODENAME main" > /etc/apt/sources.list.d/ubuntu-image.list
+# enable the ucdev PPA with additional packages to build bases
+echo "deb https://ppa.launchpadcontent.net/ucdev/base-ppa/ubuntu/ $CODENAME main" \
```
valentindavid, line 29: It is because of `https` that `ca-certificates` is needed.

alfonsosanchezbeato (commented):
Thanks for the review! I've addressed/answered the comments now.

hooks/026-configure-plymouth.chroot
alfonsosanchezbeato: Done now

static/usr/lib/systemd/system/plymouth-start.service.d/core-override.conf
alfonsosanchezbeato: Done

alfonsosanchezbeato:
I've also removed some additional packages from the plymouth deb that we don't need as suggested

valentindavid (approved):
Looks good.

static/usr/lib/systemd/system/plymouth-start.service.d/core-override.conf
```diff
@@ -0,0 +1,9 @@
+[Unit]
+After=snap-initramfs-mounts.service
```
valentindavid, line 2: This unit does not exist in core-base, and if did exist would create the same cycle as in core-initrd.

alfonsosanchezbeato: Yeah, good point. I've removed it now.

Meulengracht (approved):
LGTM!

PR #59, opened 2022-06-15
hooks: Improve robustness of fix-ld-so-symlink hook

valentindavid:
Just some pedantic changes
 - Use bash syntax
 - Disallow undefined variables
 - Do not assume there is no space in paths (though unlikely)
 - Do not assume that the symlinks have a specific path depth
 - Ignore non symlinks (still fail in test)
 - Show all errors in the test instead of exiting on first error


Meulengracht (approved):
LGTM!

PR #61, opened 2022-06-16
Refactor extra-packages hook

valentindavid:
* Use stage area to provide packages across snapcraft parts instead of
  using `$(CRAFT_PART_INSTALL)/../..`
* Make a local debian package repository
* Send all built binary packages to `extra-packages` hook. This is to
  avoid version mismatch in case some dependency start requiring other
  built packages
* Install all extra packages at once

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,16 +31,16 @@ install:
 		mknod -m 666 $(DESTDIR)/dev/urandom c 1 9
 	# copy static files verbatim
 	/bin/cp -a static/* $(DESTDIR)
-	/bin/cp $(CRAFT_PART_INSTALL)/../../probert-deb/install/*.deb $(DESTDIR)/tmp
-	/bin/cp $(CRAFT_PART_INSTALL)/../../consoleconf-deb/install/*.deb $(DESTDIR)/tmp
+	mkdir -p $(DESTDIR)/install-data
+	/bin/cp -r $(CRAFT_STAGE)/local-debs $(DESTDIR)/install-data/local-debs
 	# customize
-	set -ex; for f in ./hooks/[0-9]*.chroot; do \
-		/bin/cp -a $$f $(DESTDIR)/tmp && \
-		if ! chroot $(DESTDIR) /tmp/$$(basename $$f); then \
-                    exit 1; \
-                fi && \
-		rm -f $(DESTDIR)/tmp/$$(basename $$f); \
-	done;
+	set -eux; for f in ./hooks/[0-9]*.chroot; do		\
+		base="$$(basename "$${f}")";			\
+		cp -a "$${f}" $(DESTDIR)/install-data/;		\
+		chroot $(DESTDIR) "/install-data/$${base}";	\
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -30,16 +30,17 @@ install:
...
 	# customize
 	set -ex; for f in ./hooks/[0-9]*.chroot; do \
-		/bin/cp -a $$f $(DESTDIR)/tmp && \
-		if ! chroot $(DESTDIR) /tmp/$$(basename $$f); then \
+		/bin/cp -a $$f $(DESTDIR)/install-data/hook && \
+		if ! chroot $(DESTDIR) /install-data/hook; then \
                     exit 1; \
                 fi && \
-		rm -f $(DESTDIR)/tmp/$$(basename $$f); \
+		rm -f $(DESTDIR)/tmp/hook; \
```
alfonsosanchezbeato, line 41: Should be `rm -f $(DESTDIR)/install-data/hook;` I guess

Makefile
```diff
@@ -30,16 +30,17 @@ install:
...
 	/bin/cp -a static/* $(DESTDIR)
-	/bin/cp $(CRAFT_PART_INSTALL)/../../probert-deb/install/*.deb $(DESTDIR)/tmp
-	/bin/cp $(CRAFT_PART_INSTALL)/../../consoleconf-deb/install/*.deb $(DESTDIR)/tmp
+	mkdir -p $(DESTDIR)/install-data
+	/bin/cp -r $(CRAFT_STAGE)/local-debs $(DESTDIR)/install-data/local-debs
 	# customize
 	set -ex; for f in ./hooks/[0-9]*.chroot; do \
-		/bin/cp -a $$f $(DESTDIR)/tmp && \
-		if ! chroot $(DESTDIR) /tmp/$$(basename $$f); then \
+		/bin/cp -a $$f $(DESTDIR)/install-data/hook && \
```
alfonsosanchezbeato, line 37: Maybe a nitpick, but if you call all of the hooks "hook", `$0` won't reflect the original name, which is something sometimes useful (printing traces, etc.)

build-package.sh
```diff
@@ -0,0 +1,35 @@
...
+        source="$(dpkg-parsechangelog -SSource)"
+        version="$(dpkg-parsechangelog -SVersion)"
+        arch="$(dpkg --print-architecture)"
+        dcmd mv "../${source}_${version}_${arch}.changes" "${CRAFT_PART_INSTALL}/local-debs"
+        ;;
+    stage)
+        craftctl default
+        cd "${CRAFT_STAGE}/local-debs"
+        dpkg-scanpackages . >Packages
+        apt-ftparchive release . >Release
```
alfonsosanchezbeato, line 33: This idea of creating a local archive is very nice!

alfonsosanchezbeato (commented):
Looks very nice, I have a couple of minor comments

Makefile
valentindavid: Fixed.

alfonsosanchezbeato (approved):
LGTM, thanks for the changes

PR #62, opened 2022-06-16
Use mount namespace instead of chroot

valentindavid:
This allow to not worry about mounts done within the namespace.  We
can now bind mount files into the sysroot instead of copying them.


valentindavid:
I would like to get #61 merged first and I will rebase it. I keep it as draft for the moment.

mount-ns.sh
```diff
@@ -0,0 +1,105 @@
...
+                    esac
+                    mount --bind "${extra_args[@]}" "$2" "${sysroot}/$3"
+                    shift 3
+                    ;;
+                *)
+                    break
+                    ;;
+            esac
+        done
+        exec unshare --mount --root="${sysroot}" -- "${@}"
```
alfonsosanchezbeato, line 99: Maybe it would be good to add comments on why we need to call unshare twice (from spawn, then from init) and why we use exec the second time.

mount-ns.sh
```diff
@@ -0,0 +1,105 @@
...
+        cleanup() {
+            umount "${sysroot}/dev" || true
+            umount "${sysroot}/tmp" || true
+            umount "${sysroot}/run" || true
+        }
+        mount -t tmpfs -o mode=0755 tmpfs "${sysroot}/dev"
+        mount -t tmpfs -o mode=1777 tmpfs "${sysroot}/tmp"
+        mount -t tmpfs -o mode=0755 tmpfs "${sysroot}/run"
+        trap cleanup EXIT
+        unshare --pid --fork --mount -- "${0}" init "${sysroot}" "${@}"
```
alfonsosanchezbeato, line 59: Please add comment clarifying that you are calling the script again with a different command.

alfonsosanchezbeato (commented):
The approach is interesting but tbh I am not totally bought on this, which are the advantages from your POV?

Also, please try to build on launchpad, as maybe because of container limitations we find some problem.

mount-ns.sh
valentindavid: Done.

I have also noticed there was a potential race condition with the tmpfs when launch multiple time in the same time. So I have changed the tmpfs to be mounted in a temporary directory and bind mounted in the second phase.

valentindavid:
> The approach is interesting but tbh I am not totally bought on this, which are the advantages from your POV?

Namespaces can isolate better than chroot in general.

My main reason here is that we can hide the mounts done within the namespace to the rest system. So when the build is done and some mounts were not unmounted properly, it is fine, because they will be removed with the namespace.

Mounting of `/proc` in `hooks/001-extra-packages.chroot` was the reason I did this. First of all, we should always have `/proc` mounted. And if we did mount things in the hook scripts, then we should make sure they are not exposed to the rest of the system.

PR #64, opened 2022-06-24
snapcraft.yaml: Factorize debian package building parts

valentindavid:


xnox:
This is cute, I like it!

PR #65, opened 2022-06-24
Build initrd skeleton as part of core-base

valentindavid:


xnox:
I dislike this. My dream was to split ubuntu-core-initramfs into -bin and -$arch data files. Such that one can co-install skeletons for all arches on a given machine and generate initrd for any arch on any host. And then package ubuntu-core-initramfs as a snap.

Then when building kernel snaps, one can build-depend on core-initrd snap to generate initrd for a given kernel snap.

xnox:
Note that classic systems do not need this in general.

PR #68, opened 2022-06-29
Fix downloading of base image when mirrors are not in sync

valentindavid:
The mirrors use a round robin on `cdimage.ubunutu.com`. So calling
wget multiple times might resolve to different mirrors that might be
out of sync. Instead we need to call wget once to download all the
files.


Meulengracht (approved):
Nice, and the build is passing, good enough for me, LGTM!

PR #77, opened 2022-09-08
Trigger fix-hctosys through systemd instead of directory in udev

valentindavid:
Fixes https://bugs.launchpad.net/baoshan/+bug/1987724

### DIFF
--- a/static/usr/lib/systemd/system/fix-hctosys.service
+++ b/static/usr/lib/systemd/system/fix-hctosys.service
@@ -1,0 +1,6 @@
+[Unit]
+Description=Work around automatic hctosys
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/core/fix-hctosys
--- a/static/usr/lib/udev/rules.d/90-fix-hctosys.rules
+++ b/static/usr/lib/udev/rules.d/90-fix-hctosys.rules
@@ -1,3 +1,3 @@
 # when rtc* is added run a workaround script, see
 # https://github.com/snapcore/core20/pull/136
-ACTION=="add", SUBSYSTEM=="rtc", KERNEL=="rtc*", RUN+="/usr/lib/core/fix-hctosys"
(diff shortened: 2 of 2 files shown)


valentindavid:
Not tested though.

static/usr/lib/systemd/system/fix-htctosys.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=Work around automatic htctosys
```
mvo5, line 2: (nitpick) I would make this a bit more descriptive, maybe: `Work around time moving backwards when a rtc module is inserted and the kernel htctosys option is set` (or something like this, maybe this line is a bit too long).

mvo5 (approved):
Thank you! One small nitpick

static/usr/lib/systemd/system/fix-htctosys.service
valentindavid: I am not sure this is the point of `Description` in [systemd.unit](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Description=). It says "this string should identify the unit rather than describe it".

If it failed to run it would say in the log: "Failed to start Work around time moving backwards when a rtc module is inserted and the kernel htctosys option is set". Which is a bit weird.

Meulengracht (approved):
It looks good to me, but I haven't tested this - so if you're confident, so am i!

static/usr/lib/systemd/system/fix-htctosys.service
alfonsosanchezbeato: Nitpick: isn't it hctosys instead of htctosys? And in fact the service name has the same issue.

alfonsosanchezbeato (approved):
LGTM

static/usr/lib/systemd/system/fix-htctosys.service
valentindavid: Right!

valentindavid: Fixed.

mvo5: Thank you, I was actually not aware that description is meant this way.

PR #78, opened 2022-09-22
doc: remove reference to using the impish rootfs

IsaacJT:
As the core22 is now based on the 22.04 rootfs, this TODO can be removed.

Signed-off-by: Isaac True <isaac.true@canonical.com>


alfonsosanchezbeato (approved):
Nice catch, thank you!

PR #81, opened 2022-10-18
pam: add optional faillock

valentindavid:
When `/etc/writable/faillock.enabled` is present, then we use `pam_faillock` which can lock accounts for 900 seconds after 3 wrong password.

Can be tested with `faillock --user username`. Also use `faillock --user username --reset` instead of waiting 900 seconds.

Requires https://github.com/snapcore/snapd/pull/12278

### DIFF
--- a/hook-tests/031-faillock.test
+++ b/hook-tests/031-faillock.test
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+set -eu
+
+grep pam_faillock.so etc/pam.d/common-auth
--- a/hooks/031-faillock.chroot
+++ b/hooks/031-faillock.chroot
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+set -eu
+
+pam-auth-update --enable optional-faillock optional-faillock-preauth optional-faillock-authsucc
--- a/static/usr/lib/core/lockout-not-enabled.sh
+++ b/static/usr/lib/core/lockout-not-enabled.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+set -eu
+
+# This marker file is created by:
+# snap set system users.lockout=true
+! [ -f /etc/writable/account-locked.enable ]
--- a/static/usr/share/pam-configs/optional-lockout
+++ b/static/usr/share/pam-configs/optional-lockout
@@ -1,0 +1,7 @@
+Name: Optional lockout
+Default: yes
+Priority: 0
+Auth-Type: Primary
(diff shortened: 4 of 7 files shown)


static/usr/lib/core/lockout-not-enabled.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -eu
+
+! [ -f /etc/writable/faillock.enabled ]
```
mvo5, line 5: This now needs to be `/etc/writable/account-locked.enabled` after the latest tweaks in https://github.com/snapcore/snapd/pull/12278

hook-tests/031-faillock.test
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
mvo5, line 1: (nitick) - this could be `/bin/sh` as well, right? Nothing bash specific in here?

hooks/031-faillock.chroot
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
mvo5, line 1: (same as above)

mvo5 (approved):
Thank you! One small tweak required but this looks good. Once successfully landed we should also port it to core20 :)

hook-tests/031-faillock.test
valentindavid: Why do we want `/bin/sh`?

mvo5: There is no deep reason, it's fine (hence the "nitpick"). It's just that on Ubuntu /bin/sh is dash which is slightly faster than bash but that also does not really matter. I guess it's a bit of a habit because Debian maintainer scripts are encouraged to run with /bin/sh (because there the speed actually adds up over thousands of sh runs from the maintainer scripts) and because in theory one could have a system without bash - but this is not really relevant here and as you can see all of my arguments are rather weak :) So fine to leave if you prefer that.

Meulengracht (approved):
LGTM!

hook-tests/031-faillock.test
valentindavid: This is a very good explanation. Thank you.

static/usr/lib/core/lockout-not-enabled.sh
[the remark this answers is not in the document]
valentindavid: Good point. Done.

PR #83, opened 2022-10-20
Remove some unmounting errors on shutdown

valentindavid:
Because `/` is a squashfs from `/run/mnt/data`, we cannot unmount `/run/mnt/data`. (Or `/run/mnt/ubuntu-seed` during install/recovery).

So we need to disable `/run/mnt/data` by adding it to `shutdown.target`. However we need also to remount read-only. To avoid with writable bind mounts, we need to do it after `umount.target`.

This does not remove errors from finalrd which will be fixed in a different PR.

Note for testing: this happens after journald has stopped writing to logs. So you need to use serial port to properly log the errors.

### DIFF
--- a/static/usr/lib/systemd/system/run-mnt-data.mount.d/late-umount.conf
+++ b/static/usr/lib/systemd/system/run-mnt-data.mount.d/late-umount.conf
@@ -1,0 +1,3 @@
+[Unit]
+# This will remove implicit "Conflicts=umount.target"
+DefaultDependencies=no
--- a/static/usr/lib/systemd/system/keep-entangled-mounts.target
+++ b/static/usr/lib/systemd/system/keep-entangled-mounts.target
@@ -1,0 +1,7 @@
+[Unit]
+Description=Keep entangled mounts
+DefaultDependencies=no
+Wants=run-mnt-data.mount
+Wants=run-mnt-ubuntu\x2dseed.mount
+Wants=usr-lib-modules.mount
+Wants=run-mnt-kernel.mount
--- a/static/usr/lib/systemd/system/remount-data.service
+++ b/static/usr/lib/systemd/system/remount-data.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Remounting /run/mnt/data read-only
+DefaultDependencies=no
+ConditionPathIsMountPoint=/run/mnt/data
+After=umount.target
+Before=shutdown.target
+
+[Service]
+Type=oneshot
+ExecStart=/bin/mount -o remount,ro /run/mnt/data
--- a/static/usr/lib/systemd/system/remount-seed.service
+++ b/static/usr/lib/systemd/system/remount-seed.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Remounting /run/mnt/ubuntu-seed read-only
+DefaultDependencies=no
+ConditionPathIsMountPoint=/run/mnt/ubuntu-seed
+After=umount.target
+Before=shutdown.target
+
+[Service]
+Type=oneshot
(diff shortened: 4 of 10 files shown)


static/usr/lib/systemd/system/keep-entanlged-mounts.target
```diff
@@ -0,0 +1,4 @@
+[Unit]
+Description=Keep entangled mounts
+Wants=run-mnt-data.mount
+Wants=run-mnt-ubuntu-seed.mount
```
alfonsosanchezbeato, line 4: Afaiu this is to make sure these unmount units survive umount target? Why not put these directly as links in shutdow.target.wants?

static/usr/lib/systemd/system/run-mnt-data.mount.d/late-umount.conf
```diff
@@ -0,0 +1,2 @@
+[Unit]
+DefaultDependencies=no
```
alfonsosanchezbeato, line 2: Is this to make sure these unmount units survive umount target?

alfonsosanchezbeato (commented):
I have some comments. Another question I have is, is this expected to already remove some of the errors that are printed or that needs additionally the finalrd change that you mention? Are both things part of the solution to the unmount problem or do they solve different things?

static/usr/lib/systemd/system/keep-entanlged-mounts.target
valentindavid: They would need to be symlinks. But they are generated. I suppose we could have symlinks to `/run/systemd/transient/run-mnt-data.mount`... But they could be broken symlinks, which I do not find elegant.

static/usr/lib/systemd/system/run-mnt-data.mount.d/late-umount.conf
valentindavid: Yes, I thought I commented it. But I did not. I will add a comment.

static/usr/lib/systemd/system/keep-entanlged-mounts.target
valentindavid, line 4: There is an unescaped character here, I forgot to push the fix.

static/usr/lib/systemd/system/run-mnt-data.mount.d/late-umount.conf
valentindavid: Commented

static/usr/lib/systemd/system/keep-entanlged-mounts.target
```diff
@@ -0,0 +1,4 @@
+[Unit]
```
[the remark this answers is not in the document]
valentindavid: Fixed.

static/usr/lib/systemd/system/shutdown.target.wants/keep-entanlged-mounts.target
```diff
@@ -0,0 +1 @@
+../keep-entanlged-mounts.target
```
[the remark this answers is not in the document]
valentindavid: Fixed.

static/usr/lib/systemd/system/keep-entanlged-mounts.target
valentindavid: Fixed.

valentindavid:
I will mark this as draft. I have found issues with `systemd-fsck@.service` that I need to fix.

valentindavid:
This seems to depend on snapcore/core-initrd#120 to work correctly.

alfonsosanchezbeato (approved):
LGTM

PR #84, opened 2022-10-27
Simplify systemd configuration for console-conf

valentindavid:
Having `getty@.service` and `console-conf@.service` starting each others seem to generate errors on shut down.

```
[  115.522870] systemctl[1011]: Got message type=error sender=org.freedesktop.systemd1 destination=n/a path=n/a interface=n/a member=n/a cookie=1 reply_cookie=1 signature=s error-name=org.freedesktop.systemd1.TransactionIsDestructive error-message=Transaction for getty@tty1.service/star
[  115.524325] systemctl[1011]: t is destructive (systemd-poweroff.service has 'start' job queued, but 'stop' is included in transaction).
[  115.525067] systemctl[1011]: Failed to start getty@tty1.service: Transaction for getty@tty1.service/start is destructive (systemd-poweroff.service has 'start' job queued, but 'stop' is included in transaction).
```

Instead we should just have both `console-conf@` and `getty@` enabled and select with conditions.

Serial port services are enabled by a generator, in a similar way as `systemd-getty-generator` for `serial-getty@`.

Automatic service for virtual terminals will not for console-conf however.  This is because logind starts `autovt@.service` which is a symlink to `getty@`. To have console-conf on autovt, we would need to have `autovt@.service` start both `getty@` and `console-conf@`. However the feature of having console-conf on all VTs is pretty useless and does not justify a hack.

### DIFF
--- a/hooks/009-fix-console-conf.chroot
+++ b/hooks/009-fix-console-conf.chroot
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+set -eu
+
+rm /usr/lib/systemd/system/getty@.service.d/console-conf.conf
+
+cat <<EOF >/usr/lib/systemd/system/getty@.service.d/console-conf.conf
+[Unit]
+ConditionPathExists=/var/lib/console-conf/complete
+ConditionPathExists=!/run/snapd-recovery-chooser-triggered
+EOF
+
+rm /usr/lib/systemd/system/serial-getty@.service.d/console-conf-serial.conf
+
+cat <<EOF >/usr/lib/systemd/system/serial-getty@.service.d/console-conf-serial.conf
+[Unit]
+ConditionPathExists=/var/lib/console-conf/complete
+EOF
+
+for unit in console-conf@.service serial-console-conf@.service; do
+    mkdir -p "/usr/lib/systemd/system/${unit}.d"
+    cat <<EOF >"/usr/lib/systemd/system/${unit}.d/core.conf"
+[Unit]
+# console-conf will try to stop itself. It should not.
+RefuseManualStop=true
+[Service]
+# console-conf should try to start or stop getty@%i
+ExecStartPre=
+ExecStopPost=
+EOF
+done
+
+ln -s /usr/lib/systemd/system/console-conf@.service /etc/systemd/system/getty.target.wants/console-conf@tty1.service
--- a/static/usr/lib/systemd/system-generators/serial-console-conf
+++ b/static/usr/lib/systemd/system-generators/serial-console-conf
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+set -eux
+
+NORMAL_DIR="${1}"
+EARLY_DIR="${2}"
+LATE_DIR="${3}"
+
+for tty in $(cat /sys/class/tty/console/active); do
+    case "${tty}" in
+      tty[0-9]*)
+      ;;
+      *)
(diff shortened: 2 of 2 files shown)


valentindavid:
... It is missing the automatic restart. I will add that.

valentindavid:
There is automatic restart. But console-conf stops its service :vomiting_face: . I suppose we have to rename the units so it does not find them.

valentindavid:
It seems that this causes `console-conf@` trying to start `getty@` while `shutdown.target` is  starting is causing with some services conflicting with `shutdown.target` not being stopped. That include `finalrd.service`. That means `systemd-shutdown` is not being called in a switched root. Which means the device mapper for ubuntu-data is not taken down before powering off.

valentindavid:
I have tried to see if changing `ExecStopPost=` to `ExecStopPost=-` on `*console-conf@` did anything, but the issue still remains. So it has nothing with the exit status of `console-conf@`.

valentindavid:
> But console-conf stops its service

Pushed a work-around. We now refuse manual stop.

static/usr/lib/systemd/system-generators/serial-console-conf
```diff
@@ -0,0 +1,18 @@
...
+EARLY_DIR="${2}"
+LATE_DIR="${3}"
+
+for tty in $(cat /sys/class/tty/console/active); do
+    case "${tty}" in
+      tty[0-9]*)
+      ;;
+      *)
+        mkdir -p "${NORMAL_DIR}/getty.target.wants"
+        ln -sf /usr/lib/systemd/system/serial-console-conf@.service "${NORMAL_DIR}/getty.target.wants/serial-console-conf@${tty}.service"
```
alfonsosanchezbeato, line 15: I understand the commit message on why we do not do the same for vts, but still not sure - my worry is that maybe in the future we could have something different to console=tty1 in UC's kernel command line. Especially as we actually have tty0 usually on classic.

hooks/009-fix-console-conf.chroot
```diff
@@ -0,0 +1,33 @@
...
+[Unit]
+ConditionPathExists=/var/lib/console-conf/complete
+ConditionPathExists=!/run/snapd-recovery-chooser-triggered
+EOF
+
+rm /usr/lib/systemd/system/serial-getty@.service.d/console-conf-serial.conf
+
+cat <<EOF >/usr/lib/systemd/system/serial-getty@.service.d/console-conf-serial.conf
+[Unit]
+ConditionPathExists=/var/lib/console-conf/complete
```
alfonsosanchezbeato, line 17: Why not check also for snapd-recovery-chooser-triggered here?

alfonsosanchezbeato (commented):
Looks good, but I wonder if we should create a PR first to fix this in upstream?

hooks/009-fix-console-conf.chroot
valentindavid: Good point!

valentindavid: `serial-console-conf@.service` only has `ConditionPathExists=!/var/lib/console-conf/complete`. And we want the complement of it. So it is only `ConditionPathExists=/var/lib/console-conf/complete` we want.

valentindavid:
This is the upstream PR: canonical/subiquity#1478

valentindavid:
This was fixed by #187.

PR #85, opened 2022-11-08
Fix mount propagation

valentindavid:
To switch root, systemd has to recursively make all mounts private, then after it recursively make all mounts shared.

However `/run/mnt/*` and `/writable` are used to be bind mounted in the rest of the file system. For example, there is no reason for mount `/snap/hello/42` to also show up as
`/writable/system-data/snap/hello/42` and
`/run/mnt/data/system-data/snap/hello/42`.

### DIFF
--- a/static/usr/lib/systemd/system/remount-core-fs.service
+++ b/static/usr/lib/systemd/system/remount-core-fs.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Reset propagation of initial mount points
+DefaultDependencies=no
+Before=local-fs-pre.target
+Before=local-fs.target
+Before=shutdown.target
+Wants=local-fs-pre.target
+Conflicts=shutdown.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/lib/core/remount-core-fs
+
+[Install]
+WantedBy=local-fs.target
--- a/static/usr/lib/core/remount-core-fs
+++ b/static/usr/lib/core/remount-core-fs
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+FILESYSTEMS=(
+    /run/mnt/base
+    /run/mnt/data
+    /run/mnt/gadget
+    /run/mnt/kernel
+    /run/mnt/snapd
+    /run/mnt/ubuntu-boot
+    /run/mnt/ubuntu-save
+    /run/mnt/ubuntu-seed
+    /writable
+)
+
+for fs in "${FILESYSTEMS[@]}"; do
+    if mountpoint -q "${fs}"; then
+        mount --make-private "${fs}"
+    fi
+done
--- a/static/usr/lib/systemd/system/local-fs.target.wants/remount-core-fs.service
+++ b/static/usr/lib/systemd/system/local-fs.target.wants/remount-core-fs.service
@@ -1,0 +1,1 @@
+../remount-core-fs.service


valentindavid:
None of these path with be available in `/var/lib/snapd/hostfs`. But I think they are already not accessible.

static/usr/lib/systemd/system/remount-core-fs.service
```diff
@@ -0,0 +1,16 @@
...
+Before=shutdown.target
+Wants=local-fs-pre.target
+Conflicts=shutdown.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/lib/core/remount-core-fs
+
+[Install]
```
alfonsosanchezbeato, line 15: Install section is not needed and it looks like in general we don't add it unless it is the case.

static/usr/lib/systemd/system/remount-core-fs.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Reset propagation of initial mount points
+DefaultDependencies=no
+Before=local-fs-pre.target
+Before=local-fs.target
```
alfonsosanchezbeato, line 5: Is this needed? `local-fs-pre.target` is expected to happen before afaiu.

static/usr/lib/systemd/system/remount-core-fs.service
```diff
@@ -0,0 +1,16 @@
...
+Before=shutdown.target
+Wants=local-fs-pre.target
```
alfonsosanchezbeato, line 7: `local-fs-pre.target` is pulled already by `local-fs.target`

alfonsosanchezbeato (commented):
Some minor comments. FTR I have checked and `/var/lib/snapd/hostfs/writable` is empty in confined snaps and there is something in `/var/lib/snapd/hostfs/run/mnt/data/system-data/snap`, but just directories and symbolic links, so we should be safe for backwards compatibility there. Also, no interface gives access there at first sight.

static/usr/lib/systemd/system/remount-core-fs.service
valentindavid: No. It only has `After=`. Typically the `-pre` targets are only added if needed.

static/usr/lib/systemd/system/remount-core-fs.service
valentindavid: This is in case something happens with `local-fs-pre.target`, like something making it fail.

Note that I used roughly the same dependencies as `systemd-remount-fs.service`.

static/usr/lib/systemd/system/remount-core-fs.service
valentindavid: I know. I am not sure how I feel about the way we do it. We put everything in `/usr/lib/systemd` because `/etc/systemd/system` is writable. But we can always symlink to `/dev/null` in `/etc` to mask services installed in `/usr`. So the reason we use `/usr/lib/systemd` does not completely work.

I have added the `Install` section as indication of where it is expected to be installed. And in that way, we could one day move to another approach than what we are doing.

I can remove it.

alfonsosanchezbeato: I actually prefer it to stay as I agree it is better to have it as a hint on where the service is expected. But then we should enforce it in next changes as good practice.

static/usr/lib/systemd/system/remount-core-fs.service
alfonsosanchezbeato: I see, thanks

alfonsosanchezbeato (approved):
LGTM, thanks

I have tested and this makes a difference for `/writable/system-data/snap/`, where mounted snap files are not seen anymore. `/run/mnt/data/system-data/snap/` did not actually seem to have the issue though.

static/usr/lib/systemd/system/remount-core-fs.service
valentindavid: I have created #86 so we only use install sections and generate the symlinks.

PR #86, opened 2022-11-09
Generate symlinks from install section in units

valentindavid:
Also ensure units are enable with a preset. And move all system unit symlinks from `/etc` to `/lib` instead of few of them.


valentindavid:
I also changed some "requires" to "wants" when they do not make sense.

PR #89, opened 2022-11-16
os-release: add "ubuntu" in ID_LIKE

valentindavid:
fwupd finds the distro boot directory in EFI partition based on os-release. Because we use `ubuntu` as directory name, we need to have it in the ID_LIKE so that fwupd finds it.

### DIFF
--- a/hooks/018-set-os-release.chroot
+++ b/hooks/018-set-os-release.chroot
@@ -8,4 +8,5 @@ PRETTY_NAME="Ubuntu Core 22"
 VERSION_ID="22"
 HOME_URL="https://snapcraft.io/"
 BUG_REPORT_URL="https://bugs.launchpad.net/snappy/"
+ID_LIKE="ubuntu"
 EOF


valentindavid:
I got some questions about whether it is right to have `ID_LIKE=ubuntu`. Specially around the fact that Ubuntu Core does not have dpkg or apt.

ID_LIKE is defined by https://www.freedesktop.org/software/systemd/man/os-release.html#ID_LIKE=

> [...] It should list identifiers of operating systems that are closely related to the local operating system in regards to packaging and programming interfaces, for example listing one or more OS identifiers the local OS is a derivative from. [...] Build scripts and similar should check this variable if they need to identify the local operating system and the value of ID= is not recognized. Operating systems should be listed in order of how closely the local operating system relates to the listed ones, starting with the closest. [...]

Ubuntu Classic for example has  `ID_LIKE=debian`.

For comparison with some other image based distros based on other distros:
 * Fedora Silverblue has `ID=fedora` and `VARIANT_ID=silverblue`. It has rpm. But does not have dnf.
 * Endless OS has `ID_LIKE="ubuntu debian"`. It has dpkg and apt.
 * Steam OS 3 has `ID_LIKE=arch`. It has pacman.

xnox (changes requested):
this is not an appropriate change.

lots of things will try to indeed use dpkg / apt etc when finding this.

PR #90, opened 2022-11-23
Now build core24

valentindavid:


valentindavid:
> Waiting on [snapcore/snapd#12345](https://github.com/snapcore/snapd/pull/12345)

Was merged.

Makefile
```diff
@@ -46,9 +46,9 @@ install:
 	touch $(DESTDIR)/usr/lib/clock-epoch
 
 	# only generate manifest and dpkg.yaml files for lp build
-	if [ -e /build/core22 ]; then \
-		/bin/cp $(DESTDIR)/usr/share/snappy/dpkg.list /build/core22/core22-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).manifest; \
-		/bin/cp $(DESTDIR)/usr/share/snappy/dpkg.yaml /build/core22/core22-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).dpkg.yaml; \
+	if [ -e /build/core24 ]; then \
```
alfonsosanchezbeato, line 49: Could we have a makefile variable to avoid us repeting the current base too much?

hooks/001-extra-packages.chroot
```diff
@@ -23,13 +23,14 @@ mkdir -p /var/log/journal
...
 apt install --no-install-recommends -y ca-certificates
 
-# shellcheck disable=SC1091
-CODENAME=$(. /etc/os-release; echo "$UBUNTU_CODENAME")
-# enable the ucdev PPA with additional packages to build bases
-echo "deb https://ppa.launchpadcontent.net/ucdev/base-ppa/ubuntu/ $CODENAME main" \
-     > /etc/apt/sources.list.d/ubuntu-image.list
-
-cat >/etc/apt/trusted.gpg.d/ucdev-base-ppa.asc <<'EOF'
+if false; then
```
alfonsosanchezbeato, line 26: Why are not we doing this anymore?

snapcraft.yaml
```diff
@@ -53,9 +54,9 @@ parts:
     build-packages:
       - wget
     build-environment:
-      - LTS: jammy
-      - BASE: ${LTS}-base-${CRAFT_TARGET_ARCH}.tar.gz
-      - DIR_URL: https://cdimage.ubuntu.com/ubuntu-base/${LTS}/daily/current
+      - LTS: "22.10"
```
alfonsosanchezbeato, line 57: Nitpick: not really an LTS now, maybe this should be named differently

tests/lib/prepare-uc.sh
```diff
@@ -7,10 +7,10 @@ set -x
 . "$TESTSLIB/prepare-utils.sh"
 
 # install dependencies
-install_core22_deps
+install_core24_deps
```
alfonsosanchezbeato, line 10: Renaming to install_base_deps would be nice

tests/lib/prepare-uc.sh
```diff
@@ -7,10 +7,10 @@ set -x
...
 
 # download snaps required for us to build the image
-download_core22_snaps "$SNAP_BRANCH"
+download_core24_snaps "$SNAP_BRANCH"
```
alfonsosanchezbeato, line 13: download_base_snaps or similar

tests/lib/prepare-uc.sh
```diff
@@ -115,17 +115,17 @@ rm upstream-snapd.snap
...
+    build_core24_snap "$PROJECT_PATH"
 else
-    # use provided core22 snap
+    # use provided core24 snap
     cp "$PROJECT_PATH/core${UC_VERSION}.artifact" "$uc_snap"
 fi
 
 # finally build the uc image
-build_core22_image
+build_core24_image
```
alfonsosanchezbeato, line 128: build_base_image or similar

.github/workflows/tests.yaml
```diff
@@ -80,8 +80,9 @@ jobs:
           path: core-base
       - uses: actions/checkout@v2
         with:
-          repository: 'snapcore/snapd'
+          repository: 'valentindavid/snapd'
           path: snapd
+          ref: 'de7785a57c3517d8d08700ed638709509b1148a6'
```
alfonsosanchezbeato, line 85: This should not be needed now, right?

tests/lib/prepare-utils.sh
```diff
@@ -137,15 +137,21 @@ install_core22_deps() {
     sudo snap install ubuntu-image --classic
 }
 
-download_core22_snaps() {
+download_core24_snaps() {
     local snap_branch="$1"
 
     # get the model
-    curl -o ubuntu-core-amd64-dangerous.model https://raw.githubusercontent.com/snapcore/models/master/ubuntu-core-22-amd64-dangerous.model
+    #curl -o ubuntu-core-amd64-dangerous.model https://raw.githubusercontent.com/snapcore/models/master/ubuntu-core-24-amd64-dangerous.model
```
alfonsosanchezbeato, line 144: Maybe add a TODO here to get it from the models repo when we have the modle there?

tests/lib/prepare-utils.sh
```diff
@@ -137,15 +137,21 @@ install_core22_deps() {
...
+    cp "${PROJECT_PATH}/ubuntu-core-24-amd64-dangerous.model" ubuntu-core-amd64-dangerous.model
 
     # download neccessary images
     snap download pc-kernel --channel=22/${snap_branch} --basename=upstream-pc-kernel
     snap download pc --channel=22/${snap_branch} --basename=upstream-pc-gadget
+    unsquashfs -d upstream-pc-gadget upstream-pc-gadget.snap
```
alfonsosanchezbeato, line 150: Please add TODO for when the 24 channel is available

alfonsosanchezbeato (commented):
Looks nice! I have some comments, in general I think we should try to minimize the references to coreXX to just core/base whenever possible so this is less annoying in the future.

hooks/001-extra-packages.chroot
valentindavid: The PPA does not exist yet. It would fail.

Makefile
valentindavid: Done

tests/lib/prepare-uc.sh
valentindavid: Done.

tests/lib/prepare-uc.sh
valentindavid: Well, given the implementation is only for core24, I am not sure I want to change that.

tests/lib/prepare-uc.sh
valentindavid: Done

.github/workflows/tests.yaml
valentindavid: Removed

tests/lib/prepare-utils.sh
valentindavid: Done.

tests/lib/prepare-utils.sh
valentindavid: Done.

tests/lib/prepare-uc.sh
alfonsosanchezbeato: Ok, that's fine

hooks/001-extra-packages.chroot
Meulengracht: Would be nice to open a PR readding this so it's visible

valentindavid: Reading?

valentindavid: Ah, re-adding. Well it will be needed once we need to backport some packages. Which is probably not going to happen until we get close to or passed 24.04. It does not need to be re-added if we do not have any package.

Meulengracht: Ah yes, excuse my lack of hyphen, that definitely reads wrong. I just wanted to make sure that it's added again if neccessary, but then a TODO is probably good enough,

alfonsosanchezbeato (approved):
LGTM

PR #91, opened 2022-12-02
snapd.sshd-keygen.service: fix unit dependency to sshd

valentindavid:
The correct name is `ssh.service`


alfonsosanchezbeato (approved):
LGTM! Looking into this, at least until 22.04 there was a link to ssh.service in /etc/systemd/system/sshd.service, and actually it is there on core22 base. This must have changed in kinetic.

PR #93, opened 2023-01-05
Rework disabling of unmounts

valentindavid:
Some mounts cannot be unmounted during `shutdown.target`.  We have tried make them part of `shutdown.target` so they are kept alive. But because they depend on `system-systemd\\x2dfsck.slice` and `systemd-fsck@.service` which both conflict with `shutdown.target`, it confuses systemd, and the result of what gets scheduled is not deterministic.

Instead we remove conflicts of `system-systemd\\x2dfsck.slice` and `systemd-fsck@.service`. Then `DefaultDependencies=no` on mounts is enough to disable unmounts.

### DIFF
--- a/hooks/101-systemd-fsck.chroot
+++ b/hooks/101-systemd-fsck.chroot
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+set -eux
+
+echo "Remove conflict from systemd-fsck to shutdown"
+sed -i "/^Conflicts=shutdown.target$/d" lib/systemd/system/systemd-fsck@.service
+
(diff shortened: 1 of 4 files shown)


valentindavid:
I have proposed the change upstream in the systemd mailing list.

alfonsosanchezbeato:
I have been testing this but I see some failures on power off:
```
[   84.317424] sd-umoun[1991]: Failed to unmount /oldroot: Device or resource busy
[   84.319303] sd-umoun[1992]: Failed to unmount /oldroot/run/mnt/data: Device or resource busy
[   84.323322] sd-umoun[1993]: Failed to unmount /oldroot/run: Device or resource busy
[   84.327164] shutdown[1]: Could not detach loopback /dev/loop0: Device or resource busy
[   84.330593] shutdown[1]: Failed to finalize file systems, loop devices, ignoring.
[   84.335252] reboot: Power down
```

Is this expected, or this was actually fixing another thing?

alfonsosanchezbeato:
Great, thanks for helping me remember that. Testing does not show those errors, so LGTM.

alfonsosanchezbeato (approved):

PR #96, opened 2023-02-21
static: add generic nvidia udev rules

xnox:
These integrate nvidia graphics cards with logind and enable power management.

### DIFF
--- a/static/usr/lib/udev/rules.d/71-nvidia.rules
+++ b/static/usr/lib/udev/rules.d/71-nvidia.rules
@@ -1,0 +1,17 @@
+# 0x10de is NVIDIA vendor id
+#
+# Tag the device as master-of-seat so that logind is happy
+# (see LP: #1365336)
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
Should this maybe come from a snapd interface with a udev slot instead? cc @pedronis 

Also, is this coming from a deb?

xnox:
This will not come from a deb, because each one of them is versioned on per nvidia series, and they contain other commands/stanzas that cannot be executed on core. These are only relevant pieces that core needs.

I currently do not wish to develop a unique interface that will achieve this, as it is a system behavior, not something that individual snaps would want to request.

pedronis:
we do ship udev rules with the bases, so I'm not against this as it also relate to logind which is not shipped in a separate snap, we can revisit this for core24 as we learn more on our journey to support graphics stacks

pedronis (approved):
I'm not opposed to this as such

alfonsosanchezbeato (approved):
Thanks for the comments, I think this makes sense

PR #101, opened 2023-02-22
Update README for core24

valentindavid:
There were a few things left unchanged. Some warnings are added in things that are not yet confirmed to work. And the rest has been tested manually.


valentindavid:
After this is merged we should trigger workflow `Build LXD image`. And then update the section about LXD.

alfonsosanchezbeato (approved):
LGTM

valentindavid:
Opened #151 instead.

PR #105, opened 2023-02-27
Remove unavailable packages

valentindavid:


valentindavid:
I wonder if plymouth-label-ft is merged upstream or we need to upload it to the ppa.

alfonsosanchezbeato (changes requested):
Please keep `plymouth-label-ft`, it is required (and starting with 22.10 is built from the archive, see https://launchpad.net/ubuntu/+source/plymouth/).

alfonsosanchezbeato:
Concretely, `label-ft.so` is dynamically loaded by plymouth.

PR #112, opened 2023-03-22
Forward port some missing commits from core22 into main

alfonsosanchezbeato:


valentindavid:
"hooks: include ppa:ubuntu-security/fde-ice as part of the builds" will not work. It has to be put in the `if false; then [...] fi`

valentindavid:
I think we are in sync now.

PR #115, opened 2023-04-13
Don't include secureboot-db

vorlonofportland:
The secureboot-db package exists to update dbx at boot to revoke insecure certificates/binaries.  However, changing dbx changes the measurements in the TPM, and we seal against these measurements when doing full-disk encryption.  So we need to NOT update dbx as a result of updating the core snap, but instead need dbx updates to be managed more directly by snapd.

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -155,12 +155,6 @@ case "$(dpkg --print-architecture)" in
         ;;
 esac
 
-case "$(dpkg --print-architecture)" in
-    amd64|i386)
-        PACKAGES+=(secureboot-db)
-        ;;
-esac
-
 apt install --no-install-recommends -y "${PACKAGES[@]}"
 
 apt autoremove -y
--- a/static/usr/lib/systemd/system/core.start-snapd.service
+++ b/static/usr/lib/systemd/system/core.start-snapd.service
@@ -1,8 +1,6 @@
 [Unit]
 Description=Start the snapd services from the snapd snap
 RequiresMountsFor=/run
-Wants=secureboot-db.service
-After=secureboot-db.service
 
 [Service]
 ExecStart=/usr/lib/core/run-snapd-from-snap start


alfonsosanchezbeato:
Tests are failing here, although it seems unrelated to the PR:
```
:: /install-data/001-extra-packages.chroot: line 78: CODENAME: unbound variable
```

xnox (changes requested):
Originally when we were working on a previous revocation for the original BootHole we agreed that we will apply secureboot-db changes from the core snap, in install mode only prior to any sealings, and not apply them again as part of regular updates of the core snap. There was no good place to ship these, hence we chose to ship it in the core snap as the most logical place that can execute this unconfined.

Hence there is this snippet - https://github.com/snapcore/core-base/blob/main/static/usr/lib/systemd/system/secureboot-db.service.d/core-override.conf

```
[Unit]
ConditionKernelCommandLine=snapd_recovery_mode=install
```

The intention was, that once snapd learns to snapshot recovery/install systems, and inspect them, in theory snapd secboot would be able to assess if applying new dbx revocations from the core snap is safe and then do such operations (snapshot new recovery system, test it, seal against new recovery system and current running system, trigger secureboot-db update, reboot).

Untill all of that is done, we continue to only apply the dbx updates once during the lifecycle of Ubuntu Core installation, in the install mode. This was designed originally for the boothole revocation between @xnox @chrisccoulson @mvo5 and was documented in the boothole planning documents.

So either we should keep this as is, or also drop the included static snippet - and explain how else we are applying dbx revocations prior to first install in an Ubuntu Core lifecycle.

If we want to update secureboot-db package in Ubuntu Classic, whilst we still don't have new sealing support for the new shim we should pin older version of secureboot-db in the ubuntu-core ppas, to keep secureboot-db package in the core snap at a compatible version with the current gadgets.

xnox:
@chrisccoulson can this be rejected please?

Meulengracht:
I am closing this PR for now, please reopen should it become relevant again

PR #120, opened 2023-05-11
many: move build to mantic

alfonsosanchezbeato:


valentindavid:
The failure should have been fixed by https://github.com/snapcore/snapd/pull/12795

Not sure why this still happens. I will look into it.

alfonsosanchezbeato:
> There are few places where there is still "23.04". Maybe we can update those.

Changed now

valentindavid (approved):
Thank you

PR #121, opened 2023-05-23
static: correct the service name used in hctosys rtc fix rule

ernestl:
https://github.com/snapcore/core-base/pull/77 was created to address problem reported in https://bugs.launchpad.net/baoshan/+bug/1987724

The added udev rule refer to service file that does not exist (fix-h**t**ctosys.service)

This was discovered in core22.


mvo5 (approved):
Thank you!

valentindavid (approved):
Thank you.

PR #122, opened 2023-05-25
static: correct service name used in hctosys rtc fix rule

ernestl:
Backport fix https://github.com/snapcore/core-base/pull/121 from branch main to core22.


static/usr/lib/udev/rules.d/90-fix-hctosys.rules
alfonsosanchezbeato, line 1: I just noticed that then it would be good to change the name of the file as well, but we can do that just in main branch I think.

alfonsosanchezbeato (approved):
LGTM

static/usr/lib/udev/rules.d/90-fix-hctosys.rules
alfonsosanchezbeato: Nevermind, of course the file name is good :)

valentindavid (approved):

PR #123, opened 2023-06-13
snapcraft.yaml: update subiquity version

kubiko:
Moving to the updated subiquity version with improved detection of the run mode.


dbungert (approved):
LGTM.  This is a rev to get a newer version of the Jammy stable branch.  Looking at the diff, I think only the recent change to `console-conf-wrapper` affects core.

alfonsosanchezbeato (approved):
LGTM

PR #124, opened 2023-06-13
snapcraft.yaml: update subiquity version

kubiko:
Picking recent changes from subiquity


alfonsosanchezbeato (approved):
LGTM (see https://github.com/snapcore/core-base/pull/123)

PR #125, opened 2023-06-13
static: correct clock epoch file in script fix-hctosys

ernestl:
Correct CLOCK_EPOCH.

This change is also proposed for backporting to core22 (branch) and core20 (repository):
 - https://github.com/snapcore/core-base/pull/126
 - https://github.com/snapcore/core20/pull/147

PR #126, opened 2023-06-13
static: correct clock epoch file in script fix-hctosys

ernestl:
Correct CLOCK_EPOCH.
This change is backported from main: https://github.com/snapcore/core-base/pull/125


alfonsosanchezbeato (approved):
Nice! LGTM

PR #129, opened 2023-06-22
get-arg: fix call to printf

valentindavid:
Anything with and "A" or a "0" in the name was not parsed correctly and interpreted as a space instead.

### DIFF
--- a/static/usr/libexec/core/get-arg
+++ b/static/usr/libexec/core/get-arg
@@ -31,7 +31,7 @@ fi
 set --
 
 # We cannot use ANSI-C quoting (e.g. $'\n') in busybox-initramfs
-whitespaces="$(printf '\t\n\v\f\r \xA0')"
+whitespaces="$(printf '\t\n\v\f\r \240')"
 in_quote=no
 param=
 current="${cmdline}"


ernestl (approved):
Tested, this works.

static/usr/libexec/core/get-arg
```diff
@@ -31,7 +31,8 @@ fi
 set --
 
 # We cannot use ANSI-C quoting (e.g. $'\n') in busybox-initramfs
-whitespaces="$(printf '\t\n\v\f\r \xA0')"
+# dash's printf builtin, so we make sure does not handle \x
+whitespaces="$(/usr/bin/printf '\t\n\v\f\r \xA0')"
```
alfonsosanchezbeato, line 35: I think that hex escape is not part of the standard, you need to use octal.

`printf '\t\n\v\f\r \240`  seems to work in dash.

valentindavid: I have tested, `\240` works for all. I have changed the fix.

Meulengracht (approved):

alfonsosanchezbeato (approved):
Thanks for the change! Please remember to change the commit message too before merging.

PR #130, opened 2023-06-28
Remove pcrphase

valentindavid:
pcrphase depends on a library that we do not provide. Since we do not use PCR 11, we can drop it for now.


alfonsosanchezbeato (approved):
LGTM

PR #134, opened 2023-07-06
Do not fail when pcrphase files do not exist

valentindavid:
On s390x, systemd-pcrphase is not installed.


alfonsosanchezbeato (approved):
LGTM

PR #137, opened 2023-07-14
Create again folders in /etc/NetworkManager

alfonsosanchezbeato:
A recent cloud-init update (23.2.1) removed the only file inside /etc/NetworkManager, so the folder disappeared from the snap. However, this folder is used by the network-manager snap layouts, so we need to keep it so bind mounts on it still work.


alfonsosanchezbeato:
Will create PRs for core20/main once this is approved.

alfonsosanchezbeato:
Closing as after some testing it looks like layouts are fine even if the directory is not there anymore (it gets created on the tmpfs).

PR #141, opened 2023-07-21
hooks: add hook to remove fixup-timesyncd.conf from base image

ernestl:
Remove fixup-timesyncd.conf from the base image to ensure systemd-timesyncd.service functions correctly.

Review Notes: 
- Not yet sure on what basis the hook numbers are allocated?
- Assume such a simple hook does not require a test?
- This needs to be backported to core22 as well, not required for core20.

CONCLUSION:
This issue was caused by not snap laying around in the build directory. Abandon.

### DIFF
--- a/hooks/901-remove_timesyncd_cleanup.chroot
+++ b/hooks/901-remove_timesyncd_cleanup.chroot
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+set -e
+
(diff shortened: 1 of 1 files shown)


valentindavid:
I have built both main and core22 and I have not find this file anywhere. I think you might have tested with an older version of core22.

PR #143, opened 2023-09-29
90-rtc-sys-time-init.rules: Optimize

valentindavid:
This is cherry-pick of https://github.com/snapcore/core20/pull/156

Two optimization here:
* We filter IMPORT{cmdline}. This causes some system calls. On core20, it also reads UEFI variables. This is very slow. We used to read the cmdline on all device. Now we do it only on rtc devices.
* We stop calling /usr/bin/systemd-escape.

The first optimization is the most effective part. On google cloud, we reduce the time to cold boot devices from around 30s to 4s.


alfonsosanchezbeato (approved):
LGTM, thanks

PR #146, opened 2023-10-05
tests: use pc 24 track

valentindavid:


tests/lib/prepare-utils.sh
```diff
@@ -148,15 +148,9 @@ download_core24_snaps() {
     cp "${PROJECT_PATH}/ubuntu-core-24-amd64-dangerous.model" ubuntu-core-amd64-dangerous.model
 
     # download neccessary images
+    # TODO: publish 24 kernel
     snap download pc-kernel --channel=22/${snap_branch} --basename=upstream-pc-kernel
```
alfonsosanchezbeato, line 152: Can't we use yet the one already published in track 24?

valentindavid: I have not tested. The build present there seems old. I will open a separate PR for that.

alfonsosanchezbeato (approved):
LGTM

PR #153, opened 2023-11-09
tests: enable arm64 tests

alfonsosanchezbeato:


tests/lib/nested.sh
```diff
@@ -103,40 +109,44 @@ start_nested_core_vm_unit(){
...
+    if os.query is-pc-amd64; then
+        QEMU_BIN=qemu-system-x86_64
+        PARAM_MACHINE="-machine q35${ATTR_KVM} -global ICH9-LPC.disable_s3=1"
+        PARAM_BIOS="-drive file=/usr/share/OVMF/OVMF_CODE${VMF_CODE}.fd,if=pflash,format=raw,unit=0,readonly=on -drive file=${WORK_DIR}/image/OVMF_VARS${VMF_VARS}.fd,if=pflash,format=raw"
+        TPM_DEVICE=tpm-tis
+        cp -f "/usr/share/OVMF/OVMF_VARS${VMF_VARS}.fd" "${WORK_DIR}/image/OVMF_VARS${VMF_VARS}.fd"
+    else
+        # Assume arm64
+        QEMU_BIN=qemu-system-aarch64
+        PARAM_MACHINE="-machine virt -cpu cortex-a57"
```
valentindavid, line 132: PARAM_CPU? I see it is used other places.

tests/lib/prepare-utils.sh
```diff
@@ -114,9 +116,17 @@ start_snapd_core_vm() {
     nested_wait_for_snap_command
 }
 
+get_arch() {
+    if os.query is-pc-amd64; then
+        printf amd64
+    else
```
valentindavid, line 122: `elif os.query is-arm64` and an error in the `else`, just for it to break if we add another architecture.

valentindavid (changes requested):
Commit message "github/workflows: run for arm" should maybe use "arm64". 2 comments. Otherwise it looks good.

tests/lib/nested.sh
alfonsosanchezbeato: Good catch, changed now

tests/lib/prepare-utils.sh
alfonsosanchezbeato: Done

alfonsosanchezbeato:
@valentindavid thanks for the review, this is ready for another pass

valentindavid (approved):
Thank you

tests/lib/prepare-utils.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
 set -e
-set -x 
+set -x
 
 SSH_PORT=${SSH_PORT:-8022}
 MON_PORT=${MON_PORT:-8888}
+# For os.query and others
+PATH=$PATH:$PROJECT_PATH/tests/lib/external/snapd-testing-tools/tools/
```
sergiocazzolato, line 9: you can also either move binaries to the path in spread.yaml or to include this in the nevironment in spread.yaml.

tests/lib/prepare-utils.sh
```diff
@@ -114,9 +116,20 @@ start_snapd_core_vm() {
...
+    else
+        printf "ERROR: unsupported archtecture\n"
+        exit 1
+    fi
+}
+
 get_core_snap_name() {
     printf -v date '%(%Y%m%d)T' -1
-    echo "core24_${date}_amd64.snap"
+    echo "core24_${date}_$(get_arch).snap"
```
sergiocazzolato, line 132: I don't see where you do this for core22, I just want to make sure it is considered.

sergiocazzolato (commented):
some questions inline

Thanks for adding this change.

tests/lib/prepare-utils.sh
alfonsosanchezbeato: Good point, changed now

sergiocazzolato (approved):
Nice, I recommend to add an extra worker to the arm tests, it is taking so long to complete the execution

alfonsosanchezbeato:
Times of tests-main with one worker per system:

```
Total 2h 18m 17s
x86 44m 20s
arm64 1h 33m 45s
```

PR #154, opened 2023-11-10
Use environment variables to tell systemd to write files in /etc/writable

valentindavid:
This is a proof of concept. Do not merge.

In systemd package, we have a [patch](https://git.launchpad.net/~ubuntu-core-dev/ubuntu/+source/systemd/tree/debian/patches/debian/UBUNTU-read-only-etc-hacks.patch?h=ubuntu-noble&id=2985d314394369be965e94ba65bf8d48b88f8308) to change the path for writing symlinks/files `/etc/localtime`, `/etc/hostname`, etc.

I tried to make a different patch. Similar in principle but more generic. And submit it upstream in PR https://github.com/systemd/systemd/pull/29831

But upstream proposed instead we use environment variables instead.

This PR patches systemd to support environment variable configuration of the paths. And shows the configuration needed for core-base.

### DIFF
--- a/myfix.patch
+++ b/myfix.patch
@@ -1,0 +1,803 @@
+diff --git a/debian/patches/debian/UBUNTU-read-only-etc-hacks.patch b/debian/patches/debian/UBUNTU-read-only-etc-hacks.patch
+deleted file mode 100644
+index 6498217c9..000000000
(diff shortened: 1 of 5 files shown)


enr0n:
I didn't look at the patch in too great of detail, since that will happen more on the upstream PR, but it generally looks good to me. The drop-in configs make sense and would be a *much* nicer way to handle this.

PR #155, opened 2023-11-16
hooks: add universe to sources

alfonsosanchezbeato:
noble-base is not including universe in sources.list anymore, add it so we can still build.

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -19,6 +19,27 @@ trap 'umount /proc' EXIT
 # systemd postinst needs this
 mkdir -p /var/log/journal
 
+# Add universe to the base - needed for gnutls-bin, libengine-pkcs11-openssl,
+# libnss-extrausers, libopts25, opensc, plymouth-label-ft, gdbserver,
+# systemd-bootchart.
+# TODO this was not needed before noble, research why.
+case "$(dpkg --print-architecture)" in
+    amd64)
+        printf "
+deb http://archive.ubuntu.com/ubuntu noble universe
+deb http://archive.ubuntu.com/ubuntu noble-updates universe
+deb http://security.ubuntu.com/ubuntu noble-security universe
+" >> /etc/apt/sources.list
+        ;;
+    *)
+        printf "
+deb http://ports.ubuntu.com/ubuntu-ports noble universe
+deb http://ports.ubuntu.com/ubuntu-ports noble-updates universe
+deb http://ports.ubuntu.com/ubuntu-ports noble-security universe
+" >> /etc/apt/sources.list
+        ;;
+esac
+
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
No arm64 packages yet in http://archive.ubuntu.com/ubuntu/dists/noble/universe/

xnox (changes requested):
This is not ok, as we should not stage any packages into core24 that do not come from main. As otherwise we have to enable Pro for builds and build everything privately.

Either we need to seed these packages and have MIR done on them, or we need to drop features / migrate them elsewhere. Or somehow get agreement that we can have these from universe, and that security team will security patch them publically rather than with pro.

alfonsosanchezbeato:
> This is not ok, as we should not stage any packages into core24 that do not come from main. As otherwise we have to enable Pro for builds and build everything privately.
> 
> Either we need to seed these packages and have MIR done on them, or we need to drop features / migrate them elsewhere. Or somehow get agreement that we can have these from universe, and that security team will security patch them publically rather than with pro.

I have opened an issue about this: https://github.com/snapcore/core-base/issues/158

Yes, I agree these should be in main or dropped, and as in this cycle we want to trim down the core24 snap it is a good opportunity to do this. But meanwhile we need the base to still build, note that these packages were already in universe in core22 and previous.

valentindavid (commented):
I am not sure why this is needed. Looking at `/etc/apt/sources.list` in the chroot of the build shows that universe is already enabled. Both on core24 and core22.

sources.list wrong on arm64?

alfonsosanchezbeato:
> I am not sure why this is needed. Looking at `/etc/apt/sources.list` in the chroot of the build shows that universe is already enabled. Both on core24 and core22.
> 
> sources.list wrong on arm64?

I do not see universe at least in lxd containers that snapcraft uses to build locally the snap. Which is enough reason to include it I think, even if LP containers do have it.

valentindavid:
> I do not see universe at least in lxd containers that snapcraft uses to build locally the snap. Which is enough reason to include it I think, even if LP containers do have it.

I see it in the container. But it is not even needed there.

It is from the tarball we download in cdimage that we need it. So it should not matter if we use lxd or any other backend for snapcraft.

alfonsosanchezbeato:
> > I do not see universe at least in lxd containers that snapcraft uses to build locally the snap. Which is enough reason to include it I think, even if LP containers do have it.
> 
> I see it in the container. But it is not even needed there.
> 
> It is from the tarball we download in cdimage that we need it. So it should not matter if we use lxd or any other backend for snapcraft.

Right, I got confused. It seems that universe is now in `sources.list` from the tarball, which was not the case when I opened the PR. So it looks like the issue is already sorted out and this is not needed anymore. Anyway, it was good to trigger the discussion about packages in universe. Closing now.

PR #156, opened 2023-11-16
many: run nested tests for UC22

alfonsosanchezbeato:
Backported from https://github.com/snapcore/core-base/pull/153


sergiocazzolato:
I see this error in tests-snapd workflow -> os.query: command not found

sergiocazzolato (changes requested):
tests failing

PR #157, opened 2023-11-25
Build fixes for noble

xnox:
Also use usr/lib paths everywhere for clarity.

### DIFF
--- a/hook-tests/014-set-motd.test
+++ b/hook-tests/014-set-motd.test
@@ -6,7 +6,7 @@ grep -q "This Ubuntu Core 24 machine is a tiny" etc/motd
 
 test ! -e etc/default/motd-news
 test ! -e etc/update-motd.d/10-help-text
-test ! -e lib/systemd/system/motd-news.service
-test ! -e lib/systemd/system/motd-news.timer
+test ! -e usr/lib/systemd/system/motd-news.service
+test ! -e usr/lib/systemd/system/motd-news.timer
 
 ! grep -q unminimize etc/update-motd.d/*
--- a/hook-tests/027-remove-pcrphase.test
+++ b/hook-tests/027-remove-pcrphase.test
@@ -2,11 +2,11 @@
 
 set -eu
 
-! [ -f lib/systemd/systemd-pcrphase ]
+! [ -f usr/lib/systemd/systemd-pcrphase ]
 for f in \
-  lib/systemd/systemd-pcrphase \
-  lib/systemd/system/*/systemd-pcrphase*.service \
-  lib/systemd/system/systemd-pcrphase*.service
+  usr/lib/systemd/systemd-pcrphase \
+  usr/lib/systemd/system/*/systemd-pcrphase*.service \
+  usr/lib/systemd/system/systemd-pcrphase*.service
 do
   ! [ -f "${f}" ]
   ! [ -L "${f}" ]
--- a/hooks/027-remove-pcrphase.chroot
+++ b/hooks/027-remove-pcrphase.chroot
@@ -7,6 +7,6 @@ set -eux
 # is meant to measure signed kernel sections (see systemd-measure)
 # We are not yet using that PCR.
 
-rm -f /lib/systemd/systemd-pcrphase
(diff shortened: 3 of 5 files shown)


valentindavid:
The kernel in track 24 has stopped booting. It worked 2 weeks ago. It looks like it is hanging in the systemd stub.

xnox:
> The kernel in track 24 has stopped booting. It worked 2 weeks ago. It looks like it is hanging in the systemd stub.

yes, but i need new core24 to build new kernel snap. chicken egg problem.

and yes, kernel stub needs fixing there too.

valentindavid:
> yes, but i need new core24 to build new kernel snap. chicken egg problem.

OK, we can merge this one. I will manually test with an old kernel and approve.

I have tried to rebuild the kernel snap. But I still get hangs. It seems the systemd stub hangs when calling LoadImage (I suppose that hangs in grub). Do you already know what the problem is?

valentindavid (approved):
Thank you. We can just force-merge.

xnox:
> I have tried to rebuild the kernel snap. But I still get hangs. It seems the systemd stub hangs when calling LoadImage (I suppose that hangs in grub). Do you already know what the problem is?

not yet, but i have it reproduced locally. i will start bisect between jammy & noble systemd stubs.

valentindavid:
Actually my issue with LoadImage were due to a bug with an old version of OVMF. I cannot manage to reproduce the same exact bug as the CI locally.

valentindavid:
Now I realize, it is aarch64 that fails.

xnox:
Yes I tackling the arm64 specific failures. With just minor patches I am good at booting things on my native laptop.

jhenstridge:
For what it is worth, I ran into some console-conf problems even with this branch applied:

```
Traceback (most recent call last):
  File "/usr/bin/console-conf", line 33, in <module>
    sys.exit(load_entry_point('subiquity==0.0.5', 'console_scripts', 'console-conf-tui')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/subiquity/console_conf/cmd/tui.py", line 90, in main
    interface.run()
  File "/usr/share/subiquity/subiquitycore/tui.py", line 374, in run
    super().run()
  File "/usr/share/subiquity/subiquitycore/core.py", line 135, in run
    raise exc
  File "/usr/lib/python3.11/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/share/subiquity/subiquitycore/async_helpers.py", line 26, in _done
    fut.result()
  File "/usr/share/subiquity/subiquitycore/context.py", line 148, in decorated_async
    return await meth(self, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/subiquity/subiquitycore/controllers/network.py", line 343, in _apply_config
    await asyncio.wait_for(
  File "/usr/lib/python3.11/asyncio/tasks.py", line 489, in wait_for
    return fut.result()
           ^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/tasks.py", line 425, in wait
    raise TypeError("Passing coroutines is forbidden, use tasks explicitly.")
TypeError: Passing coroutines is forbidden, use tasks explicitly.
```

I could patch that with the following:

```
sed 's/\(e.wait()\)\( for e in dhcp_events\)/asyncio.create_task(\1)\2/' -i subiquitycore/controllers/network.py
```

This got me as far as user creation, which failed when `adduser` tried to add my new user account to the `users` group. This seems to be a behaviour change in adduser:

https://salsa.debian.org/debian/adduser/-/blob/master/debian/NEWS#L70-78

Meulengracht (approved):
Thank you, and thanks to @jhenstridge for pointing out that additional things are required. I'm currently also doing some build changes so i'll just keep this PR in mind, but merging these changes for now

xnox:
Subiquity stuff should be raised against subiquity repo and escalated to foundations

PR #161, opened 2024-01-02
hooks: enable polkit for unconfined processes

alfonsosanchezbeato:
polkit is needed in some cases for communication between services shipped in the base, for instance when systemd-networkd talks to systemd-hostnamed to set the hostname. Enable to fix these use cases. Note that this does not help yet if we want to use it from a snap.

### DIFF
--- a/hooks/000-provide-uids-gids.chroot
+++ b/hooks/000-provide-uids-gids.chroot
@@ -115,6 +115,7 @@ _apt:x:117:65534:Reserved:/nonexistent:/bin/false
 syslog:x:108:114:Reserved:/home/syslog:/bin/false
 dnsmasq:x:109:65534:Reserved:/var/lib/misc:/bin/false
 tss:x:110:116:Reserved:/var/lib/tpm:/bin/false
+polkitd:x:111:120:polkit:/nonexistent:/usr/sbin/nologin
 EOF
 cp /etc/passwd /etc/passwd.orig # We make a copy for a later sanity-compare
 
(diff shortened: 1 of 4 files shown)


valentindavid (commented):
Two comments so far.

* xml-core is installed. I do not think it makes sense to keep it. It believe it is only useful for sgml docs. I am guessing it is a dependency of the post install script of some of the packages.
* `/usr/lib/tmpfiles.d/polkitd.conf` points to `/etc/polkit-1/rules.d` and `/var/lib/polkit-1`. We should add those in writable-paths. The first one is probably not needed now. But I am guessing the second should be.

valentindavid:
>     * `/usr/lib/tmpfiles.d/polkitd.conf` points to `/etc/polkit-1/rules.d` and `/var/lib/polkit-1`. We should add those in writable-paths. The first one is probably not needed now. But I am guessing the second should be.

Nevermind for  `/var/lib/polkit-1`, it seems was useful for localauthority for 3rd party packages. And it has been removed in recent version of polkit. I do not think we need it, and I do not think we every will.

alfonsosanchezbeato:
> Two comments so far.
> 
> * xml-core is installed. I do not think it makes sense to keep it. It believe it is only useful for sgml docs. I am guessing it is a dependency of the post install script of some of the packages.

It is used to generate xml catalog files, which I'm not sure if polkit xml parser will need. I have removed the executables provided by xml-core, as we definitely don't need them after installation of packages has happened.

> * `/usr/lib/tmpfiles.d/polkitd.conf` points to `/etc/polkit-1/rules.d` and `/var/lib/polkit-1`. We should add those in writable-paths. The first one is probably not needed now. But I am guessing the second should be.

Goot point, I have added the one under /etc as it would be something we will eventually want for snaps or for workarounds.

alfonsosanchezbeato:
@valentindavid thanks for your review, I've addressed your comments now

valentindavid (approved):

PR #162, opened 2024-01-02
Update subiquity and add console-conf fixes

alfonsosanchezbeato:


alfonsosanchezbeato:
Opening as draft for the moment as there are some changes that might be actually upstream bugs.

alfonsosanchezbeato:
https://github.com/canonical/subiquity/pull/1881 should fix the subiquitycore dependency.

alfonsosanchezbeato:
The change that adds the "users" group to the extrausers database is needed because in noble the `adduser` script adds by default new users to said group: https://git.launchpad.net/ubuntu/+source/adduser/tree/adduser?h=applied/ubuntu/noble#n654

This could be overridden by changing `/etc/adduser.conf`.

alfonsosanchezbeato:
Changing to ready for review as the dependency is now added in subiquity core package.

valentindavid (approved):
It seems to me to be a bug for extrausers not being able to add a user to an existing group.

I approve this work-around. Because it does not look like extrausers is actively maintained.

But we should really move to systemd-homed. It can add users to a group from /etc/group.

PR #165, opened 2024-01-16
66-snapd-autoimport.rules: filter out non filesystem blocks

valentindavid:
Because we might run `snap auto-import --mount` on a partition table block and its partition blocks at the same time, it might cause a conflict.  We then get an error like `/dev/sda1 already mounted or mount point busy.`.

To go around this issue, we trigger the service only for devices that are identified as filesystems.


alfonsosanchezbeato (approved):
LGTM

Meulengracht (approved):
lgtm

PR #166, opened 2024-01-16
many: convert core24 to build using chisel

Meulengracht:
Update/add/remove hooks to reflect this. We no longer depend on the ubuntu-base image to build, instead pull everything we need directly from chisel. Everything unfortunately must go together as none of these changes can really stand on their own.

Below is the raw data of the conversion:

Files added:
https://paste.ubuntu.com/p/qQJ6XCbf7f/

Files removed:
https://paste.ubuntu.com/p/qjmSSs6yJ5/

Original snap size:    77.5MB
Chiselled snap size:  80.4MB


hooks/610-cleanup-xml-core.chroot
```diff
@@ -1,14 +1,10 @@
-#!/bin/sh
+#!/bin/sh -ex
 #
 # Remove parts of the xml-core package - runnable bits are used by polkit
 # package postinst/prerm scripts and it is not needed after that.
 
-set -ex
```
valentindavid, line 6: I prefer using `set` than in the shebang. The reason is if something calls "sh ./the-script.sh" instead of "./the-script.sh", the parameters in the shebang are lost.

Meulengracht: Will update the hooks to reflect this style, and thank you for making me aware of this

hooks/000-provide-uids-gids.chroot
```diff
@@ -12,67 +12,67 @@
...
-news:x:9:9:news:/var/spool/news:/usr/sbin/nologin
-uucp:x:10:10:uucp:/var/spool/uucp:/usr/sbin/nologin
-proxy:x:13:13:proxy:/bin:/usr/sbin/nologin
-www-data:x:33:33:www-data:/var/www:/usr/sbin/nologin
-backup:x:34:34:backup:/var/backups:/usr/sbin/nologin
-list:x:38:38:Mailing List Manager:/var/list:/usr/sbin/nologin
-irc:x:39:39:ircd:/run/ircd:/usr/sbin/nologin
-_apt:x:42:65534::/nonexistent:/usr/sbin/nologin
-nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
+root:*:0:0:root:/root:/bin/bash
```
valentindavid, line 15: I am not sure this is normal to use `*` here, unless you removed them from `/etc/shadow`. But I do not think that would be correct either.

snapcraft.yaml
```diff
@@ -12,6 +12,144 @@ grade: devel
 assumes: [snapd2.55.5]
 
 parts:
+  chisel-libs:
+    plugin: nil
+    build-packages:
+      - golang
+    source: https://github.com/Meulengracht/chisel-releases.git
```
valentindavid, line 19: How do you enable PPAs? We usually enable https://launchpad.net/~ucdev/+archive/ubuntu/base-ppa But for now on core24, it is not needed. But it probably will be. Is that in `chisel.yaml` there?

Meulengracht: I will owe you an answer on how this works. Since packages are fetched in a building context in snapcraft, I guess we simply add it the way you would normally do to enable PPAs. Snapcraft has built-in syntax for this?

hooks/000-provide-uids-gids.chroot
Meulengracht: The reason I changed this is because the `/etc/passwd` that now comes from the deb package through chisel looks like this (it's merely a copy of `/usr/share/base-passwd/group.master`). Which means this diff check fails here. I looked up the `*` modifier and it means (from manpages)

```
If the password field contains some string that is not a valid result
of crypt(3), for instance ! or *, the user will not be able to use a
unix password to log in (but the user may log in the system by other
means).
```

valentindavid: Right. It is not where we want to say that. We want `*` that in `/etc/shadow`. We need `x` in `/etc/passwd` to say we are using `/etc/shadow`. Unprivileged users do not have to know if a password has been set for the users. `*` gives too much information.

snapcraft.yaml
valentindavid: Do you mean that chisel takes the host apt configuration from `/etc/apt`?

Meulengracht: PPAs for chisel are not supported currently: https://github.com/canonical/chisel/issues/98

From the code it looks like the PPA url is hardcoded: https://github.com/canonical/chisel/blob/a88e10523947ae15e1f87116b56803e13d3324b4/internal/archive/archive.go#L129

I'll bring this up with the chisel team

hooks/000-provide-uids-gids.chroot
Meulengracht: I'll mark this as resolved as we write our own /etc/* files

Makefile
```diff
@@ -47,11 +45,11 @@ install:
 	# see https://github.com/systemd/systemd/blob/v247/src/shared/clock-util.c#L145
 	touch $(DESTDIR)/usr/lib/clock-epoch
 
-	# only generate manifest and dpkg.yaml files for lp build
-	if [ -e $(BUILDDIR) ]; then \
-		/bin/cp $(DESTDIR)/usr/share/snappy/dpkg.list $(BUILDDIR)/$(SNAP_NAME)-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).manifest; \
-		/bin/cp $(DESTDIR)/usr/share/snappy/dpkg.yaml $(BUILDDIR)/$(SNAP_NAME)-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).dpkg.yaml; \
-	fi;
+	# TODO: Chisel can't generate a list of packages yet.
```
alfonsosanchezbeato, line 48: How would a manifest that would be used to detect CVEs be generated?

PR #168, opened 2024-01-18
Remove systemd-pcrmachine.service

valentindavid:
This runs `systemd-pcrphase` which we do not use and we have removed.


Meulengracht (approved):
lgtm

alfonsosanchezbeato (approved):
LGTM, thanks

PR #172, opened 2024-01-24
66-snapd-autoimport.rules: only import from removable disks

valentindavid:


alfonsosanchezbeato (commented):
I think this is fine, but this needs some double check from field as maybe we have put user assertions in the past in normal partitions of a non-removable disk for some reason, so I've added @kubiko as a reviewer.

valentindavid:
@alfonsosanchezbeato I suggest we do not backport that. But it would be nice to be in core24. My understanding is that we documented it only for "removable USB storage device".

Meulengracht (approved):
Thank you, lets await @kubiko 's approval

alfonsosanchezbeato:
> @alfonsosanchezbeato I suggest we do not backport that. But it would be nice to be in core24. My understanding is that we documented it only for "removable USB storage device".

Yes, +1 on not backporting

kubiko (approved):
+1 from me, very neat. Thank you @valentindavid!
As for backport, I am not opposed to backport, as this was always advertised as something only functioning with external media, a prime use-case USB pen drive. I believe there was even an effort on UC20 or so to explicitly exclude internal drives.

alfonsosanchezbeato (approved):
LGTM, thanks!

PR #173, opened 2024-01-24
snapd.autoimport-device@.service: do not run along snapd.autoimport.service

valentindavid:
There is a conflict between snapd.autoimport.service and snapd.autoimport-device@.service. This causes error messages because both cannot mount in the same time. So they cannot be run at the same time.


Meulengracht (approved):
Thank you, is there other versions of core this fix is relevant for?

alfonsosanchezbeato (approved):
LGTM, thanks - not sure if maybe we should have somewhat a single service for both already mounted or not mounted, but well, something to think about in the future...

PR #174, opened 2024-01-25
snapd.autoimport-device@.service: do not run along snapd.autoimport.service

valentindavid:
Backport of #173 

There is a conflict between snapd.autoimport.service and snapd.autoimport-device@.service. This causes error messages because both cannot mount in the same time. So they cannot be run at the same time.


alfonsosanchezbeato (approved):
LGTM

PR #175, opened 2024-02-01
snapcraft: drop unnecessary files (part 1)

bboozzoo:
Drop a bunch of files that are not useful at runtime, this includes:
- stripped down docs
- lintian
- plain text hwdb entries
- gdb plugins and support files

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -101,3 +101,23 @@ parts:
       craftctl default
       # ensure build-in tests are run
       cd ${CRAFT_PART_SRC} && make test TESTDIR=${CRAFT_PRIME}
+    stage:
+      # curated list of things we can drop to save space
+
+      # docs
+      - -usr/share/doc/*
+      - -usr/share/doc-base
+      # lintian overrides
+      - -usr/share/lintian
+      # valgrind suppress profiles for python
+      - -usr/lib/valgrind
+      # remove individual hwdb files, runtime only uses the binary database
+      - -usr/lib/udev/hwdb
+      # gdb auto-load plugins, but there is no gdb in the base, just gdbserver
+      - -usr/share/gdb
+      # support for gdb plugin for untangling libstdc++ containers
+      - -usr/share/gcc/python
+
+      # TODO perl
+      # TODO adduser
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -101,3 +101,23 @@ parts:
       craftctl default
       # ensure build-in tests are run
       cd ${CRAFT_PART_SRC} && make test TESTDIR=${CRAFT_PRIME}
+    stage:
+      # curated list of things we can drop to save space
+
+      # docs
+      - -usr/share/doc/*
```
valentindavid, line 108: I think there are copyright notices in there that might need to stay.

snapcraft.yaml
```diff
@@ -101,3 +101,23 @@ parts:
...
+      # remove individual hwdb files, runtime only uses the binary database
+      - -usr/lib/udev/hwdb
+      # gdb auto-load plugins, but there is no gdb in the base, just gdbserver
+      - -usr/share/gdb
+      # support for gdb plugin for untangling libstdc++ containers
+      - -usr/share/gcc/python
+
+      # TODO perl
+      # TODO adduser
+      # TODO X11 xkb
```
Meulengracht, line 123: I have a slight suspecion we might remove these in a hook, and not in the stage list, so I'm unsure these TODOs need to be here

snapcraft.yaml
```diff
@@ -101,3 +101,23 @@ parts:
       craftctl default
       # ensure build-in tests are run
       cd ${CRAFT_PART_SRC} && make test TESTDIR=${CRAFT_PRIME}
+    stage:
```
Meulengracht, line 104: While I like that we can use stage for these things, it's not consistent with how we usually remove stuff from the base. We usually remove stuff in the hooks, and it'll be confusing if things are split. Maybe @alfonsosanchezbeato has an opinion on this

snapcraft.yaml
bboozzoo: I can move all of it into a hook if that's a better choice

bboozzoo:
I'll move to hooks and open a new PR

PR #176, opened 2024-02-02
hooks: drop non-essential files

bboozzoo:
A bunch of cleanups of files that appear to be unnecessary and do not need to be shipped in core24.

AMD64 core24 size goes down by 1604kB.

### DIFF
--- a/hooks/603-cleanup-docs.chroot
+++ b/hooks/603-cleanup-docs.chroot
@@ -1,0 +1,16 @@
+#!/bin/sh
+#
+# Prune unnecessary documentation, leave the copyright bits
+
+set -ex
+
+echo "I: Removing docs"
+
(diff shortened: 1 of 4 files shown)


alfonsosanchezbeato (commented):
Nice! Some additional things that I think could be removed:
* /usr/share/bug
* /usr/share/cmake
* /usr/share/apport
* /usr/share/debhelper
* /usr/share/debianutils
* /usr/share/zsh
and maybe others in /usr/share.

bboozzoo:
Also added info files (since we don't ship info anyway) on top of the additional entries suggested @alfonsosanchezbeato

Meulengracht (approved):
This looks great and very muich in line with a lot of the removal I did during chisel

alfonsosanchezbeato (approved):
LGTM, thanks!

PR #181, opened 2024-02-16
hooks: remove file, libmagic and its data

bboozzoo:
file is added as a dependendency of cracklib-runtime. However we already drop all the binary programos of cracklib-runtime package and only leave the cache files, hence file becomes non-essential. The libmagic dependency of file pulls in libmagic-mgc carrying a 8MB+ binary file with file identification data.

Saves ~500kB on the snap size.

### DIFF
--- a/hooks/040-remove-pkgs.chroot
+++ b/hooks/040-remove-pkgs.chroot
@@ -4,3 +4,8 @@ set -e
 
 echo "Remove uneeded packages"
 dpkg -r --force-depends dh-python
+
+
+# Remove file which got pulled in by cracklib-runtime. We are discarding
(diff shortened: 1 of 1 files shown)


valentindavid (approved):
Thank you.

* I have verified that cracklib-runtime is the only dependency that pulls file and libmagic.
* I have tried to check what in cracklib-runtime depends on file. But I think it does not need it. It might have been true a long time ago. Cracklib does provide a magic file eventually, but I do not think it is provided on Ubuntu anywhere. I think it is just an old dependency that never got removed.

Meulengracht (approved):
Thank you

PR #184, opened 2024-02-19
run-snapd-from-snap: set the current symlink before bootstrapping

valentindavid:
Needed by https://github.com/snapcore/snapd/pull/13370

### DIFF
--- a/static/usr/lib/core/run-snapd-from-snap
+++ b/static/usr/lib/core/run-snapd-from-snap
@@ -9,6 +9,9 @@ set -eux
 run_on_unseeded() {
     SNAPD_BASE_DIR="/run/mnt/snapd"
 
+    [ -d /snap/snapd ] || mkdir -p /snap/snapd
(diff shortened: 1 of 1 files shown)


static/usr/lib/core/run-snapd-from-snap
```diff
@@ -9,6 +9,9 @@ set -eux
 run_on_unseeded() {
     SNAPD_BASE_DIR="/run/mnt/snapd"
 
+    [ -d /snap/snapd ] || mkdir -p /snap/snapd
+    ln -sf "${SNAPD_BASE_DIR}" /snap/snapd/current
```
alfonsosanchezbeato, line 13: Why is this needed now and not in the past? Maybe this should have a comment.

valentindavid: I have added a comment.

alfonsosanchezbeato (approved):
LGTM, thanks!

PR #186, opened 2024-02-20
hooks: remove dhcpcd-base

Meulengracht:
Cloud-init now pulls in dhcpcd-base, which we are not interested in.

### DIFF
--- a/hooks/040-remove-pkgs.chroot
+++ b/hooks/040-remove-pkgs.chroot
@@ -9,3 +9,10 @@ dpkg -r --force-depends dh-python
 # Remove file which got pulled in by cracklib-runtime. We are discarding
 # cracklib-runtime binary programs in later hooks (601-clean-cracklib.chroot).
 dpkg -r --force-depends file libmagic1 libmagic-mgc
+
+# cloud-init pulls in dhcpcd-base which we aren't (currently)
+# interested in. It's marked as required, but cloud-init works with
+# multiple dhcp clients. This change occurred recently (from time of commit)
+# since isc-dhcp-client will be deprecated and is currently in the process
+# of being demoted from packages in main universe.
+dpkg --purge --force-depends dhcpcd-base


hooks/040-remove-pkgs.chroot
```diff
@@ -9,3 +9,10 @@ dpkg -r --force-depends dh-python
 # Remove file which got pulled in by cracklib-runtime. We are discarding
 # cracklib-runtime binary programs in later hooks (601-clean-cracklib.chroot).
 dpkg -r --force-depends file libmagic1 libmagic-mgc
+
+# cloud-init pulls in dhcpcd-base which we aren't (currently)
+# interested in. It's marked as required, but cloud-init works with
+# multiple dhcp clients. This change occurred recently (from time of commit)
+# since isc-dhcp-client will be deprecated and is currently in the process
+# of being demoted from packages in main universe.
+dpkg --purge --force-depends dhcpcd-base
```
alfonsosanchezbeato, line 18: nitpick: maybe a bug should be opened against cloud-init so they improve the packaging to make this optional, and then add the reference to the bug here.

alfonsosanchezbeato (approved):
LGTM but please look at comment

Meulengracht:
Tested that should anyone need dhcdpcd, they can include it in their snap and run dhcpcd as root. The user created by the package is not required for operating.

holmanb:
> However cloud-init supports multiple dhcp clients and does not in fact need dhcpcd-base currently.

@Meulengracht true, but cloud-init still requires a working dhcp client to be installed

> nitpick: maybe a bug should be opened against cloud-init so they improve the packaging to make this optional, and then add the reference to the bug here.

@alfonsosanchezbeato Good point, but this was already considered. The Ubuntu deb packaging requires only dhcpcd-base because dhcpcd-base is expected to be the only client in main[1] on Noble that meets the needs of cloud-init. Since isc-dhcp-client isn't expected to remain in main in 24.04, there is no point as including it as an optional dependency.

Using the isc-dhcp-client is only expected on Ubuntu in the following cases 1) Ubuntu on series older than Noble and 2) Noble+ if the user manually installs isc-dhcp-client and then forces cloud-init to use it via a configuration setting. Since dhcpcd-base is included in the initramfs, ubuntu-minimal, etc[2], it is expected that on Noble and later, dhcpcd-base will always be available.

[1] 2 if you include busybox, but that is [currently not fully supported](https://github.com/canonical/cloud-init/issues/4765) yet

[2] various projects that made this transition off of isc-dhcp-client
```
$ apt rdepends dhcpcd-base
dhcpcd-base
Reverse Depends:
  Depends: cloud-init
 |Recommends: madwimax
 |Depends: libguestfs0
  Depends: dhcpcd
  Depends: ubuntu-minimal
  Depends: initramfs-tools-core
```

PR #192, opened 2024-03-15
hooks: create xdg-email as a copy of xdg-open

soumyaDghosh:
Many apps, specifically browsers like chromium, brave doesn't work with xdg-open only to open mailto links. This would help those apps to be able to open mailto links.
Also can revert this commit, https://git.launchpad.net/review-tools/commit/?id=31c894178a8fd110b9f65f0ee532b7e11384f8b8

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -23,10 +23,11 @@ mkdir -p /var/log/journal
 apt update
 apt install --no-install-recommends -y ca-certificates
 
+# shellcheck disable=SC1091
+CODENAME=$(. /etc/os-release; echo "$UBUNTU_CODENAME")
+
 # TODO: There is no PPA for noble yet.
 if false; then
-    # shellcheck disable=SC1091
-    CODENAME=$(. /etc/os-release; echo "$UBUNTU_CODENAME")
     # enable the ucdev PPA with additional packages to build bases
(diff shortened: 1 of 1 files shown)


hooks/500-create-xdg.chroot
```diff
@@ -29,6 +29,8 @@ x-scheme-handler/mailto=xdg-open.desktop
 x-scheme-handler/help=xdg-open.desktop
 EOF
 
+cp /usr/bin/xdg-open /usr/bin/xdg-email
```
Meulengracht, line 32: could it have been a symlink?

alfonsosanchezbeato: Also, it does not look like they are the same binary in fact, not in noble.

alfonsosanchezbeato: Oh, but I see `xdg-open` is a script we create in this hook nevermind.

alfonsosanchezbeato (commented):
Probably this is fine, but please create a symlink instead and do it nearer to where we write `/usr/bin/xdg-open`.

alfonsosanchezbeato:
Also, please amend commit message so it is `hooks:...` instead of `feat:...`

soumyaDghosh:
Sorry for wrong commit message, recreating this PR. Please forgive me for the inconveniences.

PR #199, opened 2024-03-22
Add systemd-coredump

valentindavid:
Easy way to fix https://forum.snapcraft.io/t/core-dumps-in-snapd/920


valentindavid:
Quick way to test:
```
sleep 1000 &
pkill -ABRT sleep
sudo coredumpctl dump --output=core
```

alfonsosanchezbeato (commented):
Note that the main issue is to make sure that core dumps are produced by confined snaps, would this be enough for that? Also, we need a system option to enable/disable the dumps.

alfonsosanchezbeato:
Also we need to consider if it makes sense to make things compatible with `apport`.

alfonsosanchezbeato:
Closing as this was merged as part of another PR.

PR #200, opened 2024-04-02
static/usr/lib/tmpfiles.d: create journald config drop-in directory

Meulengracht:
@bboozzoo as per the suggestion in MM.

This would allow us to be able to grant access to the journald.conf.d directory, allowing customers to create drop-ins and customize the journald experience.

I'll open similar ones for core22/core20 if we agree.

Per @valentindavid suggestion, use tmpfiles.d to create the directory as it won't be created otherwise because of how /etc/systemd is mounted in writable paths.


bboozzoo (approved):
LGTM

valentindavid (approved):

bboozzoo (approved):

PR #203, opened 2024-04-08
emergency: wait for 10 seconds before rebooting

valentindavid:
This is a backport of https://github.com/snapcore/core-initrd/pull/124

Will "undraft" it after I have tested it. I opened it to remember.

### DIFF
--- a/static/usr/lib/systemd/system/emergency.service.d/core-override.conf
+++ b/static/usr/lib/systemd/system/emergency.service.d/core-override.conf
@@ -1,2 +1,2 @@
 [Unit]
-AssertKernelCommandLine=dangerous
+ConditionKernelCommandLine=dangerous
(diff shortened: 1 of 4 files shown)


static/usr/lib/systemd/system/emergency.service.d/core-override.conf
```diff
@@ -1,2 +1,2 @@
 [Unit]
-AssertKernelCommandLine=dangerous
+ConditionKernelCommandLine=dangerous
```
Meulengracht, line 2: This is indeed much nicer

Meulengracht (approved):
The units look good - but how come this is needed in the base if it's in the initramfs?

valentindavid:
> The units look good - but how come this is needed in the base if it's in the initramfs?

This can still happen in base part of the boot. If an important mount does not work for example, we will have the emergency started. Or anything that has `OnFailure=emergency.target`.

PR #208, opened 2024-04-10
hooks: update motd

Meulengracht:
Update the motd of Ubuntu Core - and remove the default 00-header motd that used to print 'Welcome to ...'

It's easier to have full control of the MOTD, instead of having the welcome message as well, which is simply sourced from /etc/lsb-release, and doesn't even mention 'core'.


alfonsosanchezbeato (approved):
LGTM, thanks

PR #209, opened 2024-04-11
Use ubuntu-image stable channel

valentindavid:


alfonsosanchezbeato (approved):
LGTM - cross fingers on tests passing

valentindavid:
ssh user is not created in some tests. I have added a work-around for cloud-init. Not sure if that helps.

PR #211, opened 2024-04-15
tests: work-around cloud-init bug

valentindavid:


Meulengracht (approved):
Thank you for porting that over!

PR #212, opened 2024-04-15
tests: use beta branch with edge model

valentindavid:
The edge model uses the pc-kernel beta. So if we use edge, it will automatically update on first boot.

### DIFF
--- a/tests/lib/prepare-utils.sh
+++ b/tests/lib/prepare-utils.sh
@@ -151,14 +151,26 @@ install_base_deps() {
 }
 
 download_core24_snaps() {
+    # FIXME: there is no reason to select a branch when the model is
+    # hard coded.
     local snap_branch="$1"
 
     # get the model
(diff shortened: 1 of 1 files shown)


tests/lib/prepare-utils.sh
```diff
@@ -158,7 +158,15 @@ download_core24_snaps() {
 
     # download neccessary images
     # TODO: publish 24 kernel
-    snap download pc-kernel --channel=24/${snap_branch} --basename=upstream-pc-kernel
+    case "${snap_branch}" in
```
Meulengracht, line 161: this warrants a comment

Meulengracht (approved):
Thanks for fixing it!!

tests/lib/prepare-utils.sh
valentindavid: Added a comment. Removed the TODO. Added a fixme.

PR #216, opened 2024-04-22
hooks: do not remove dhcpcd-base

Meulengracht:
It seems that not having a dhcpcd-base client causes issues with usages of cloud-init. We see this on GCE and in Certifications testing when using the images.

As long as cloud-init is a part of the base, it makes better sense to also contain its dependencies.

For more info see:
https://github.com/snapcore/core-base/pull/186#issuecomment-2018299100
https://github.com/canonical/cloud-init/issues/5088

This can be causing the issue we are seeing in UC related to failing to create users through adduser - as it may be falling back to creating through that if it cannot otherwise resolve other data-sources.


alfonsosanchezbeato (approved):
LGTM, thank you

PR #217, opened 2024-04-23
hooks: override default groups by cloud-init to only mention extrauser groups

Meulengracht:


alfonsosanchezbeato (commented):
I think that adding this to `10_snappy.cfg` should work too:
```
system_info:
   default_user:
     groups: []
```

hooks/032-cloud-init.chroot
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+set -eu
+
+# add override the default groups to lxd, as the rest cannot be modified on ubuntu-core
+sed -i '/\ \ \ \ \ lock_passwd: True/a\ \ \ \ \ groups: [lxd]' etc/cloud/cloud.cfg.d/10_snappy.cfg
```
valentindavid, line 6: I am surprised it works with `[lxd]`. I would have just set `[]`.

alfonsosanchezbeato:
FWIW if I try to add any group to `groups` it does not work, includng lxd group.

alfonsosanchezbeato:
Just noticed you were using `sed` in the snappy conf (got confused about that), actually you can edit directly `/static/etc/cloud/cloud.cfg.d/10_snappy.cfg`

alfonsosanchezbeato (approved):
LGTM, thanks!

PR #218, opened 2024-04-24
Update 001-extra-packages.chroot for 64-bit time_t

ginggs:
Replace libp11-3 with libp11-3t64 in PACKAGES

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -145,7 +145,7 @@ PACKAGES=(
     libengine-pkcs11-openssl
     libnss-extrausers
     libopts25
-    libp11-3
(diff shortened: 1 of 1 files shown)


Meulengracht:
To add a bit of context, this PR is to fix the build for armhf

alfonsosanchezbeato (approved):
LGTM, thanks

Meulengracht:
To add some reasoning from Lukasz:
> We had to remove a bunch of armhf binaries for the time_t transition. Normally we'd rebuild and make sure to either change all the package dependencies to the t64 ones or add a Provides: to make sure the old ones still work, but there was a rush at the end. It was just faster/safer to remove the armhf binaries sometimes.

PR #219, opened 2024-04-24
hooks: create locale configuration 

alfonsosanchezbeato:
This is read by pam and others.

### DIFF
--- a/hooks/009-locale-archive.chroot
+++ b/hooks/009-locale-archive.chroot
@@ -24,3 +24,7 @@ for encoding in "${!found_encodings[@]}"; do
 done
 
 apt purge -y locales
+
+# restore configuration so the default locale can be found
+printf 'LANG=C.UTF-8\n' > /etc/locale.conf
+ln -s ../locale.conf /etc/default/locale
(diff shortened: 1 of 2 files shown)


Meulengracht (approved):

alfonsosanchezbeato:
Please don't merge this yet as I do not see that the files are actually appearing in the snap.

alfonsosanchezbeato:
Removing blocked label after changing the approach

hooks/009-locale-archive.chroot
```diff
@@ -23,4 +23,12 @@ for encoding in "${!found_encodings[@]}"; do
...
 done
 
+# Backup configuration
+cp /etc/locale.conf /etc/locale.conf.bak
+
+# purge
 apt purge -y locales
+
+# and restore so the default locale can be found
+mv /etc/locale.conf.bak /etc/locale.conf
```
valentindavid, line 33: Could we just put LANG=C.UTF-8, without taking it from the image=

valentindavid: Also it needs to be writable in theory. So we should add it to the writable-paths

alfonsosanchezbeato: I've changed this now

valentindavid (approved):
Thank you.

PR #220, opened 2024-04-25
10_snappy.cfg: use correct value type to silence a deprecation from cloud…

Meulengracht:
After speaking to the server team it seems we produce an DEPRECATED message due to the schema validation. Silence this to avoid this message.

This partially (another thing is also causing this, see https://github.com/canonical/cloud-init/issues/5192) causes `cloud-init status` to return an error code '2', even though it produces the output `status: done`.

This causes snapd to treat it like an error, even though there was no error.

The fix we actually need is to make sure snapd checks for `status: done` instead of relying on the error code


alfonsosanchezbeato (approved):
LGTM

valentindavid (approved):
Yaml...

PR #224, opened 2024-05-28
snapcraft: switch to release image

Meulengracht:
The LP build was broken since the daily images are now ocular. Switch to the correct release images


alfonsosanchezbeato (approved):
LGTM, thanks

PR #226, opened 2024-06-13
bump subiquity and probert

mwhudson:
This bumps probert and subiquity/console-conf to commits that have been relicensed to GPL3-only.

I have built the core22 snap locally with and without this change and the differences are only what is expected.

As the comment change suggests, I have created branched in the subiquity and probert repos to track the commits that go into core22, we could change core22 to follow those branches if that would make things smoother (and you trust us not to mess with them, I guess).

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -18,8 +18,8 @@ parts:
       - devscripts
     source: https://github.com/canonical/probert.git
     source-type: git
-    # following branch server/jammy
-    source-commit: 253acbe1480fba69d12d6e72f53412da79e27410
+    # following branch core/jammy
+    source-commit: 94be2fa3f45e766ecee526fdc0c15f65a78667e6
     override-pull: |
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato (commented):
This should be proposed first to `main`

mwhudson:
console-conf isn't part of core24 though?

alfonsosanchezbeato:
> console-conf isn't part of core24 though?

Oh, indeed :slightly_smiling_face: . @bboozzoo do we need to update to these commits in the console-conf snap?

alfonsosanchezbeato (approved):
LGTM, thank you

bboozzoo:
> > console-conf isn't part of core24 though?
> 
> Oh, indeed 🙂 . @bboozzoo do we need to update to these commits in the console-conf snap?

IDK, what is the context for this?

Meulengracht (approved):
Thanks!

PR #229, opened 2024-07-19
run-snapd-from-snap: seed with access to keyring

valentindavid:
The new FDE manager needs access the kernel keyring to access the primary key. That includes seeding.

### DIFF
--- a/static/usr/lib/core/run-snapd-from-snap
+++ b/static/usr/lib/core/run-snapd-from-snap
@@ -21,7 +21,7 @@ run_on_unseeded() {
     # a systemd socket unit so that systemd own the socket, otherwise
     # the socket file would be removed by snapd on exit and the snapd.seeded
     # service will fail because it has nothing to talk to anymore.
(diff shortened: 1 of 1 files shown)


valentindavid:
@pedronis In my tests, it seems I need to access the keyring during seeding. It is needed to mark the first boot as successful. So we will need this change on all bases since core20. Unless we disallow this during seeding.

alfonsosanchezbeato (commented):
This is fine, but iiuc this implies that next snapd will depend on the base being modern enough to have this change. Not sure if a check/workaround of some sort can/should be implemented in snapd.

pedronis:
> This is fine, but iiuc this implies that next snapd will depend on the base being modern enough to have this change. Not sure if a check/workaround of some sort can/should be implemented in snapd.

given that this affects first boot, it's mostly about new images as people shouldn't really build new images with new snapd but old base?  as long as we make sure that we released this for the relevant bases before we release a snapd based on the new fde-manager-features branch and snapd fails cleanly, we should be ok

alfonsosanchezbeato (approved):
I had actually not noticed this is just for seeding, so it is fine.

PR #234, opened 2024-08-15
snapcraft,make(core24): add script for generating changelogs, include as part of build

Meulengracht:
Generates a changelog like we do for system-snaps, just between the dpkg.yaml manifests that we already generate.

The changelog is installed into `/usr/share/doc/ChangeLog` (however this is up to dicussion if you feel like we should do it somewhere else).

This needs to be backported to other core bases once it's approved

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,11 +46,31 @@ install:
 	# see https://github.com/systemd/systemd/blob/v247/src/shared/clock-util.c#L145
 	touch $(DESTDIR)/usr/lib/clock-epoch
 
+	# generate the changelog, for this we need the previous core snap
+	# to be installed, this should be handled in snapcraft.yaml
+	if [ -e "/snap/$(SNAP_NAME)/current/usr/share/snappy/dpkg.yaml" ]; then \
+		./tools/generate-changelog.py \
+			"/snap/$(SNAP_NAME)/current/usr/share/snappy/dpkg.yaml" \
+			"$(DESTDIR)/usr/share/snappy/dpkg.yaml" \
+			"$(DESTDIR)/usr/share/doc" \
+			$(DESTDIR)/usr/share/doc/ChangeLog; \
+	else \
+		echo "WARNING: changelog will not be generated for this build"; \
+	fi
+
 	# only generate manifest and dpkg.yaml files for lp build
 	if [ -e $(BUILDDIR) ]; then \
 		/bin/cp $(DESTDIR)/usr/share/snappy/dpkg.list $(BUILDDIR)/$(SNAP_NAME)-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).manifest; \
 		/bin/cp $(DESTDIR)/usr/share/snappy/dpkg.yaml $(BUILDDIR)/$(SNAP_NAME)-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).dpkg.yaml; \
+		if [ -e $(DESTDIR)/usr/share/doc/ChangeLog ]; then \
+			/bin/cp $(DESTDIR)/usr/share/doc/ChangeLog $(BUILDDIR)/$(SNAP_NAME)-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).ChangeLog; \
+		fi \
 	fi;
+
+	# after generating changelogs we can cleanup those bits
+	# from the base
+	find "$(DESTDIR)/usr/share/doc/" -name 'changelog.Debian.gz' -print -delete
+	find "$(DESTDIR)/usr/share/doc/" -name 'changelog.gz' -print -delete
 
 .PHONY: check
 check:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -55,6 +55,15 @@ parts:
     build-packages:
       - shellcheck
       - distro-info
+      - python3-debian
+      - python3-requests
+      - python3-yaml
+    # to generate changelogs, it's own snap need to be present,
+    # obviously for new core bases this cannot be the case, and
+    # we need to wait until the first release with introducing
+    # changelogs.
+    build-snaps:
+      - core24=latest/beta
     override-pull: |
       craftctl set version="$(/bin/date +%Y%m%d)"
       craftctl default
--- a/tools/generate-changelog.py
+++ b/tools/generate-changelog.py
@@ -1,0 +1,196 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2024 Canonical Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Script is originally copied and modified from
+# https://github.com/snapcore/system-snaps-cicd-tools/blob/0148d51ac8d1668f8ebe142d91673a65e520aa53/workflows/changelog-from-manifest.py
+
+# Script that compares two manifest.yaml files and creates a paragraph
+# that includes the changes for all deb files. These changes are
+# obtained from the debian changelog for the different packages.
+
+import argparse
+import debian.changelog
+import debian.debian_support
+import gzip
+import os
+import requests
+import sys
+import yaml
+from collections import namedtuple
+
+# the packages here have for some weird reason "legit"
+# reasons for not having a working changelog, thus we
+# allow download for the packages mentioned here.
+# keep the list short to not increase the time it takes
+# to generate changelogs
+pkg_allowed_list = [
+    'dpkg',   # is removed during hook
+    'openssl' # contains a symlink which is broken currently
+]
+
+# Returns a dictionary from package name to version, using
+# the packages section.
+# manifest_p: path to manifest to load
+def packages_from_manifest(manifest_p):
+    with open(manifest_p) as manifest:
+        manifest_y = yaml.safe_load(manifest)
+
+        pkg_dict = {}
+        for pkg in manifest_y['packages']:
+            # package_name=version
+            pkg_data = pkg.split('=')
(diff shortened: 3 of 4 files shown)


snapcraft.yaml
```diff
@@ -55,6 +55,8 @@ parts:
     build-packages:
       - shellcheck
       - distro-info
+    build-snaps:
```
Meulengracht, line 58: Obviously would have had bootstrap issues. For core26 we will have a slightly different solution anyway due to the fact of it being chiselled.

tools/generate-changelog.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2019 Canonical Ltd
```
valentindavid, line 3: Was this copied from somewhere?

Meulengracht: yes - it's a modified copy of the one from the system snaps CICD repo

Meulengracht: but good point about the copyright year

valentindavid: But what is the strategy here. if we fix anything here. Should we also fix it somewhere else?

tools/generate-changelog.py
```diff
@@ -0,0 +1,167 @@
...
+    if argv is None:
+        argv = sys.argv[1:]
+
+    if len(argv) != 3:
+        eprint('Usage:', sys.argv[0],
+               '<old_manifest> <new_manifest> <docs_dir>')
+        return 1
+    old_manifest = argv[0]
+    new_manifest = argv[1]
+    docs_dir = argv[2]
```
valentindavid, line 154: It is so much better to use argparse.

tools/generate-changelog.py
```diff
@@ -0,0 +1,167 @@
...
+
+# Gets difference in changelog between old and new versions
+# Returns source package and the differences
+def get_changes_for_version(docs_d, pkg, old_v, new_v, indent):
+    # Try to get changelog from file (only option that will work for
+    # ESM packages), otherwise go to changelogs.ubuntu.com.
+    try:
+        changelog = get_changelog_from_file(docs_d, pkg)
+    except Exception:
+        changelog = get_changelog_from_url(pkg, new_v)
```
valentindavid, line 82: Do we really need that? I am not a fan of downloading anything outside of the "pull" phase in snapcraft.

tools/generate-changelog.py
valentindavid: Please keep the 2019 as well, if it was modified.

Meulengracht: The strategy here is to just copy it per repo for now - once we spend some time to streamline our setup to make the base repos use the same CICD repo as the system-snaps, we will consolidate the support for changelogs there.

The script is small, relatively simple and specialized, it does (IMHO) not make sense to create a new repo for just that *or* pull in the entire CICD repo to have a variant of the script there.

Meulengracht: If you or  @alfonsosanchezbeato feel strongly about the alternative (put it into cicd and pull it in), i will happily change it

valentindavid: I do not care that much. But it would be nice to say where it originally came from. Just because if it stops working, we can see if the other place already has some fix.

valentindavid: And it would be nice also to say the hash of the commit it was you copied from.

tools/generate-changelog.py
Meulengracht: good point - updated!

tools/generate-changelog.py
Meulengracht: @valentindavid I added a link to the original script and the hash, i think its good suggestions

tools/generate-changelog.py
```diff
@@ -0,0 +1,167 @@
...
+    new_packages = packages_from_manifest(new_manifest_p)
+    changes = ''
+
+    src_pkgs = {}
+    SrcPkgData = namedtuple('SrcPkgData', 'old_v new_v changes debs')
+    for pkg, new_v in sorted(new_packages.items()):
+        try:
+            old_v = old_packages[pkg]
+            if old_v != new_v:
+                (src, pkg_change) = get_changes_for_version(docs_d, pkg, old_v,
```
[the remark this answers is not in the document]
Meulengracht: updated!

tools/generate-changelog.py
Meulengracht: I do not know how some packages end of not having their changelog in the form of `changelog.Debian.gz` - but unfortunately it happens and the url function is hit - and correctly resolves the missing changelogs, so I'd prefer we keep it for now

tools/generate-changelog.py
```diff
@@ -0,0 +1,165 @@
...
+    parser = argparse.ArgumentParser(description="Manifest changelog generator")
+
+    parser.add_argument('old-manifest', help='Path to the manifest of the previous snap')
+    parser.add_argument('new-manifest', help='Path to the manifest of the new snap')
+    parser.add_argument('docs-dir', help='Path to the usr/share/doc directory in the rootfs of the new snap')
+    args = parser.parse_args()
+
+    old_manifest = getattr(args, 'old-manifest')
+    new_manifest = getattr(args, 'new-manifest')
+    docs_dir = getattr(args, 'docs-dir')
```
valentindavid, line 152: If you really want to use '-' in the help, maybe you should consider using `metavar`. For example `parser.add_argument('old', metavar='old-manifest')`. And then you can use `args.old`, instead of `old_manifest = getattr(args, 'old-manifest')`. (or even use `_` in the field if you like long names).

tools/generate-changelog.py
```diff
@@ -0,0 +1,165 @@
...
+    chl = debian.changelog.Changelog(changelog)
+    old_deb_v = debian.debian_support.Version(old_v)
+    for version in chl.get_versions():
+        vc = debian.debian_support.version_compare(old_deb_v, version)
+        if vc >= 0:
+            break
+
+    # Get the changelog chunk since the version older or equal to old_v
+    change_chunk = ''
+    old_change_start = source_pkg + ' (' + version.__str__() + ')'
```
valentindavid, line 93: I do not like the `__str__`. I would use `old_change_start = f'{source_pkg} ({version!s})'`.

tools/generate-changelog.py
```diff
@@ -0,0 +1,165 @@
...
+    for line in changelog.splitlines():
+        if line.startswith(old_change_start):
```
valentindavid, line 95: Should not we fail if we do never match it? Do not we risk to have huge changelogs if for some reason the formatting changes.

tools/generate-changelog.py
Meulengracht: We know why it falls back here now - it's because we clean up in docs in the hooks :-)

tools/generate-changelog.py
Meulengracht: added, good point!

tools/generate-changelog.py
Meulengracht: yea, switched to interpolated string, nice suggestion

tools/generate-changelog.py
Meulengracht: updated this, did not know about `metavar`, nice!

valentindavid (approved):
Thank you!

Makefile
```diff
@@ -46,12 +46,30 @@ install:
 	# see https://github.com/systemd/systemd/blob/v247/src/shared/clock-util.c#L145
 	touch $(DESTDIR)/usr/lib/clock-epoch
 
+	# generate the changelog, for this we need the previous core snap
+	# to be installed, this should be handled in snapcraft.yaml
+	if [ -e "/snap/$(SNAP_NAME)/current/usr/share/snappy/dpkg.yaml" ]; then \
```
alfonsosanchezbeato, line 51: It could be good to show a warning to the terminal in case this is not the case

alfonsosanchezbeato (commented):
Thank! I have some minor comments. Also, this should be proposed to main first (maybe we branched a bit too early)

tools/generate-changelog.py
```diff
@@ -0,0 +1,195 @@
...
+
+# the packages here have for some weird reason "legit"
+# reasons for not having a working changelog, thus we
+# allow download for the packages mentioned here.
+# keep the list short to not increase the time it takes
+# to generate changelogs
+pkg_allowed_list = [
+    'dpkg',
+    'openssl'
+]
```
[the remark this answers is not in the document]
Meulengracht: The changelog openssll here is a symlink  - and dpkg do not have a changelog present, it must have been removed in one of the previous hooks but i cannot figure out which one removed it, nothing refers to it. I'll add notes for each

snapcraft.yaml
```diff
@@ -55,6 +55,15 @@ parts:
...
       - distro-info
+      - python3-debian
+      - python3-requests
+      - python3-yaml
+    # to generate changelogs, it's own snap need to be present,
+    # obviously for new core bases this cannot be the case, and
+    # we need to wait until the first release with introducing
+    # changelogs.
+    build-snaps:
+      - core24
```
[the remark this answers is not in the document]
Meulengracht: Yea it means latest/stable by default -- I do not think it's a good idea to use beta, if we publish something to beta that does not pass beta validation, then we get a changelog that would be incorrect

tools/generate-changelog.py
```diff
@@ -0,0 +1,195 @@
...
+
+# Gets difference in changelog between old and new versions
+# Returns source package and the differences
+def get_changes_for_version(docs_d, pkg, old_v, new_v, indent):
+    # Try to get changelog from file (only option that will work for
+    # ESM packages), otherwise go to changelogs.ubuntu.com.
+    try:
+        changelog = get_changelog_from_file(docs_d, pkg)
+    except Exception:
+        changelog = get_changelog_from_url(pkg, new_v)
```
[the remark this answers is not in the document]
Meulengracht: good point, updated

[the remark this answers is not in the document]
Meulengracht: good point, updated

Makefile
Meulengracht: nice idea, added

snapcraft.yaml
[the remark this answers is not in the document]
alfonsosanchezbeato: But if it passes (which is actually the usual case), then the changelog will be incorrect too as the changes for actually 2 released revisions will be shown as having happened in just one revision.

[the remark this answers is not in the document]
Meulengracht: I changed it to beta

alfonsosanchezbeato (approved):
LGTM, thanks for the changes

PR #235, opened 2024-08-21
snapcraft,make(core22): add script for generating changelogs, include as part of build

Meulengracht:
Backport of https://github.com/canonical/core-base/pull/234


Meulengracht:
Created SNAPDENG-31648 to investigate why CI fails for core22

alfonsosanchezbeato:
Hm, tests-snapd seems to be failing because console-conf is active.

PR #237, opened 2024-08-22
feat(dhcpcd): add /var/lib/dhcpcd writable-paths for datasource disco…

blackboxsw:
…very

If dhcpcd is installed in a cloud image, cloud-init may attempt early datasource discovery using dhcpcd in order to get information about active instance metadata service IP addresses and routes. The dhcpcd client writes out leases which cloud-init parses under /var/lib/dhcpcd.

Add this support so cloud-init can detect network-bound datasources such as Azure, Ec2 and GCE.


alfonsosanchezbeato:
Forward port from https://github.com/canonical/core-base/pull/228

PR #238, opened 2024-08-22
feat(azure): support cloud-init persistent writes to /var/lib/waagent

blackboxsw:
cloud-init persists any configuration data crawled from Azure instance metadata service  or ovf-env.xml files to /var/lib/waagent directory to allow walinux-agent to interact with various early boot configuration seeds. If that directory does not exist or is readonly cloud-init will hit a permissions error resulting in inability to detect the Azure instance metadata service to obtain boot-time cloud-init configuration.

### DIFF
--- a/static/etc/system-image/writable-paths
+++ b/static/etc/system-image/writable-paths
@@ -57,6 +57,7 @@
 # cloud-init
 /etc/cloud                              auto                    persistent  transition  none
 /var/lib/cloud                          auto                    persistent  none        none
+/var/lib/waagent                        auto                    persistent  none  none
 # for various clouds like GCE
 /etc/sysctl.d                           auto                    persistent  transition  none
 # swapfile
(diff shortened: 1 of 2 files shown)


static/etc/system-image/writable-paths
```diff
@@ -57,6 +57,7 @@
 # cloud-init
 /etc/cloud                              auto                    persistent  transition  none
 /var/lib/cloud                          auto                    persistent  none        none
+/var/lib/waagent                        auto                    persistent  transition  none
```
alfonsosanchezbeato, line 60: Action should be `none`, not transition

alfonsosanchezbeato (commented):
I think that the approach is, one comment though

static/etc/system-image/writable-paths
blackboxsw: Thanks @alfonsosanchezbeato squash merged and pushed that change as a single commit.

alfonsosanchezbeato (approved):
LGTM, thanks

PR #240, opened 2024-08-30
static: wait for kernel driver for services that load modules

alfonsosanchezbeato:
Wait for kernel modules to be mounted in modprobe@.service and mount for /proc/sys/fs/binfmt_misc/, as these load modules. This is in principle a workaround until we start to mount modules and firmware from the initramfs.

### DIFF
--- a/static/usr/lib/systemd/system/modprobe@.service.d/wait-modules.conf
+++ b/static/usr/lib/systemd/system/modprobe@.service.d/wait-modules.conf
@@ -1,0 +1,3 @@
+# TODO remove once we start to mount from initramfs
+[Unit]
+After=usr-lib-modules.mount usr-lib-firmware.mount
--- a/static/usr/lib/systemd/system/proc-sys-fs-binfmt_misc.mount.d/wait-modules.conf
+++ b/static/usr/lib/systemd/system/proc-sys-fs-binfmt_misc.mount.d/wait-modules.conf
@@ -1,0 +1,3 @@
+# TODO remove once we start to mount from initramfs
+[Unit]
+After=usr-lib-modules.mount usr-lib-firmware.mount


static/usr/lib/systemd/system/modprobe@.service.d/wait-modules.conf
```diff
@@ -0,0 +1,2 @@
+[Unit]
```
valentindavid, line 1: I do not see `modeprobe@binfmt-misc.service` used in any unit. Is this really used?

static/usr/lib/systemd/system/proc-sys-fs-binfmt_misc.mount.d/wait-modules.conf
```diff
@@ -0,0 +1,2 @@
+[Unit]
```
valentindavid, line 1: If this is a temporary fix, I think we should add it as comment in the file.

valentindavid (changes requested):
A few comments.

static/usr/lib/systemd/system/modprobe@.service.d/wait-modules.conf
alfonsosanchezbeato: This is used by some units that have dependencies of they: `Wants=modprobe@<module_name>.service`, like systemd-pstore.service or sys-kernel-config.mount. It is not related to binfmt-misc.

static/usr/lib/systemd/system/proc-sys-fs-binfmt_misc.mount.d/wait-modules.conf
alfonsosanchezbeato: I've added comments to the two files now

Meulengracht (approved):

alfonsosanchezbeato:
This should not be needed anymore after changes in initramfs that mount the drivers tree from there.

PR #242, opened 2024-09-03
snapcraft.yaml: update to 24.04.1 point release

alfonsosanchezbeato:


alfonsosanchezbeato:
This is still failing due to the lxd change that did not allow mounting /proc anymore. The lxd snap in latest/candidate as of 03/09/2024 is working though.

Meulengracht (approved):
Tested locally, works

Meulengracht:
@alfonsosanchezbeato please propose this for the core24 branch as well

sergio-costas (approved):
I tested it, and works fine.

snapcraft.yaml
```diff
@@ -18,7 +18,8 @@ parts:
       - wget
     build-environment:
       - RELEASE: "24.04"
-      - BASE: ubuntu-base-${RELEASE}-base-${CRAFT_ARCH_BUILD_FOR}.tar.gz
+      - DOT_RELASE: "1"
+      - BASE: ubuntu-base-${RELEASE}.${DOT_RELASE}-base-${CRAFT_ARCH_BUILD_FOR}.tar.gz
```
sergio-costas, line 22: Maybe add this using `BASE: ubuntu-base-${RELEASE}${DOT_RELEASE:+.$DOT_RELEASE}-base-${CRAFT_ARCH_BUILD_FOR}.tar.gz` ? That way, if DOT_RELEASE isn't defined, it will use only the RELEASE version, and

snapcraft.yaml
```diff
@@ -18,7 +18,8 @@ parts:
       - wget
     build-environment:
       - RELEASE: "24.04"
-      - BASE: ubuntu-base-${RELEASE}-base-${CRAFT_ARCH_BUILD_FOR}.tar.gz
+      - DOT_RELASE: "1"
```
sergio-costas, line 21: Mmm... `dot_relEase`... shouldn't it?

sergio-costas (changes requested):

snapcraft.yaml
sergio-costas: (and the same in the next line, of course)

snapcraft.yaml
alfonsosanchezbeato: Good point, I've changed now (actually to `${DOT_RELEASE+.$DOT_RELEASE}`, I don't think you need the colon). I was not sure if snapcraft would process this as expected but seems to be the case.

snapcraft.yaml
alfonsosanchezbeato: Fixed, good catch!

snapcraft.yaml
alfonsosanchezbeato: I've rethought about the colon and put back.

sergio-costas: I tested it and yes, the colon is needed.

sergio-costas (approved):
With the last changes, looks perfect.

PR #246, opened 2024-09-09
many: introduce changelog for the snap repository itself prior to package changes

Meulengracht:
Example of changelog with this change: https://paste.ubuntu.com/p/qcyYNVwMXz/

### DIFF
--- a/tools/generate-changelog.py
+++ b/tools/generate-changelog.py
@@ -196,8 +196,16 @@ def read_remote_git_url() -> str:
     return remote_url.removesuffix(".git")
 
 
-def log_between_commits(start, end):
-    return subprocess.check_output(['git', 'log', f'{start}..{end}']).decode()
+def log_between_commits(name, start, end):
+    try:
+        return subprocess.check_output(['git', 'shortlog', '--pretty=short', f'{start}..{end}']).decode()
+    except:
+        # if there is no path from start..end then this might fail, however this
+        # should only happen if the branch has diverged so much that the previous
+        # release commit does not exist in the current fork. In this case let us
+        # notify that we could not generate the changelog
+        print(f"Failed to run 'git log' for the current repo starting at commit {start}, has branch diverged to much?")
+        return f'No detected changes for the {name} snap\n\n'
 
 
 def main():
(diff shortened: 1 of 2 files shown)


Meulengracht:
I need to fix the CI build which throws an error :-)

tools/generate-changelog.py
```diff
@@ -182,30 +196,47 @@ def read_remote_git_url() -> str:
...
     now = datetime.now()
-    changes = f"{now.strftime("%d/%m/%Y")}, commit {read_remote_git_url()}/tree/{read_commit_hash()}\n\n"
+    changes = f"{now.strftime("%d/%m/%Y")}, commit {read_remote_git_url()}/tree/{ccommit}\n\n"
+    changes += f'[ Changes in the {args.name} snap ]\n\n'
+
+    if pcommit != "":
+        changes += log_between_commits(pcommit, ccommit)
+        changes += "\n\n"
+    else:
+        changes += f'No changes for the core{args.name} snap\n\n'
```
alfonsosanchezbeato, line 231: But if pcommit==ccommit we would not have changes either, right? In fact if pcommit=="" what might be happening is that there is no previous changelog?

tools/generate-changelog.py
```diff
@@ -182,30 +196,47 @@ def read_remote_git_url() -> str:
     return remote_url.removesuffix(".git")
 
 
+def log_between_commits(start, end):
+    return subprocess.check_output(['git', 'log', f'{start}..{end}']).decode()
```
alfonsosanchezbeato, line 200: I think we should take only the title of the commit, having the full commit information is too noisy I think

alfonsosanchezbeato (commented):
Thanks! Couple of comments.

tools/generate-changelog.py
Meulengracht: How about we use `git shortlog` instead?

tools/generate-changelog.py
Meulengracht: Yea that is a nice catch, thank you, i'll just test some more cases here

tools/generate-changelog.py
Meulengracht: Now it looks like this: https://paste.ubuntu.com/p/qcyYNVwMXz/

Meulengracht: Agreed, something we need to be aware of when merging, that commits look nice

alfonsosanchezbeato (approved):
LGTM, thanks for the changes

PR #247, opened 2024-09-10
many: introduce changelog for the snap repository itself prior to package changes

Meulengracht:
Backport of https://github.com/canonical/core-base/pull/246 for core22

REF: SNAPDENG-31649

### DIFF
--- a/tools/generate-changelog.py
+++ b/tools/generate-changelog.py
@@ -39,7 +39,8 @@ from collections import namedtuple
 # keep the list short to not increase the time it takes
 # to generate changelogs
 pkg_allowed_list = [
-    'distro-info-data', # is removed during hook
+    'apt', # is removed during hook
+    'libapt-pkg6.0', # is removed as well
 ]
 
 # Returns a dictionary from package name to version, using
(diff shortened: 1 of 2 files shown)


tools/generate-changelog.py
```diff
@@ -37,7 +39,8 @@
 # keep the list short to not increase the time it takes
 # to generate changelogs
 pkg_allowed_list = [
-    'distro-info-data', # is removed during hook
```
Meulengracht, line 40: i belive this was broken before on the CI, but for some reason wasn't hit

alfonsosanchezbeato: ok, it is a bit strange though that the list here is quite different to the one for core24 (we have there dpkg, openssl, and base-files).

alfonsosanchezbeato (approved):
LGTM, thanks

tools/generate-changelog.py
Meulengracht: I already added to my TODO to look more into this - I have a sneaking suspecion we might be not removing the packages we expect since this happens

PR #249, opened 2024-09-11
snapcraft: add support for core24 fips build

Meulengracht:
To test this, rename the fips.conf.example file to fips.conf and replace your access information from LP (LOGIN and PASSWORD)

In addition, to trigger the application of the FIPS patches, you need to rename your remote for `origin` to contain the word `fips`.

REF: SNAPDENG-31750


snapcraft.yaml
```diff
@@ -61,6 +61,13 @@ parts:
       - python3-yaml
     override-pull: |
       craftctl set version="$(/bin/date +%Y%m%d)"
+      craftctl default
+    override-build: |
+      # apply fips patch if the remote is a fips remote
```
Meulengracht, line 66: i'm taking ideas on how to improve this check, it's very crude right now

snap-patch/01-switch-to-fips.patch
```diff
@@ -0,0 +1,55 @@
+diff --git a/Makefile b/Makefile
+index 769e060..fe6ea22 100644
+--- a/Makefile
++++ b/Makefile
+@@ -24,6 +24,9 @@ install:
+ 		cp /etc/apt/trusted.gpg $(DESTDIR)/etc/apt/ || true; \
+ 		cp -r /etc/apt/trusted.gpg.d $(DESTDIR)/etc/apt/ || true; \
+ 	fi
++	# Enable updates
++	sed -n 's/noble-security/noble-updates/p' /etc/apt/sources.list >> $(DESTDIR)/etc/apt/sources.list
```
valentindavid, line 10: Is not this needed only on launchpad?

snapcraft.yaml
```diff
@@ -61,6 +61,13 @@ parts:
...
+      if git remote get-url origin | grep "fips"; then
+        git apply "$CRAFT_PROJECT_DIR"/snap-patch/01-switch-to-fips.patch
```
valentindavid, line 68: I am not sure using a patch is the best thing. Because one has to update the patch any time a line it in the context is modified. Could we use an environment variable?

Also I feel like checking the remote should be done in override-pull, then write down the what mode to build in a file.

valentindavid:
> In that patch there is an identifier `<access-token>` that you need to replace with your personal access token to the FIPS Modules Under Certification key available on LP under your account.

If feel like one should just add a file in `static/etc/apt/auth.conf.d/`then remove that file in some hook.

Makefile
```diff
@@ -3,6 +3,14 @@ TESTDIR ?= "prime/"
 SNAP_NAME=core24
 BUILDDIR=/build/$(SNAP_NAME)
 
+# include any fips environmental setup if the file exists.
+# Variables:
+# - SNAP_FIPS_BUILD
+ifneq (,$(wildcard ./.fips-env))
+    include .fips-env
```
valentindavid, line 10: Instead of using a if, maybe just `-include`, with the dash as prefix.

Makefile
```diff
@@ -3,6 +3,14 @@ TESTDIR ?= "prime/"
...
+    export
```
valentindavid, line 11: I would explicitly list the variables that need export.

static/etc/apt/auth.conf.d/01-fips.conf.example
```diff
@@ -0,0 +1,3 @@
+# replace the <LOGIN> and <PASSWORD> with your own private access
```
valentindavid, line 1: Do comments work on netrc syntax? Or maybe that only works for apt. Which is a nice thing. No warnings?

snapcraft.yaml
```diff
@@ -60,8 +60,14 @@ parts:
...
       - python3-yaml
     override-pull: |
-      craftctl set version="$(/bin/date +%Y%m%d)"
       craftctl default
+      # detect whether we are doing a fips build on LP
+      if git remote get-url origin | grep "fips"; then
+        craftctl set version="$(/bin/date +%Y%m%d)+fips"
+        echo "SNAP_FIPS_BUILD=1" > ./.fips-env
+      else
+        craftctl set version="$(/bin/date +%Y%m%d)"
```
valentindavid, line 69: Not important. But I feel we should remove `.fips-env` in here. Or even make it empty. The only case I think of however is if one changes the "origin" remote. So I am being picky for not much.

I wonder, could someone just add that file manually to force a fips build? Maybe this is what you had in mind on how to build locally the fips build, unfortunately, it does not add the +fips in that case.

static/etc/apt/auth.conf.d/01-fips.conf.example
Meulengracht: Actually good question, I just assumed here - I didn't have the comments while testing

hooks/001-extra-packages.chroot
```diff
@@ -118,10 +117,65 @@ FSbrQ9ACQFlqN49Ogbl47J6TZ7BrjDpROote55ixmrU=
 -----END PGP PUBLIC KEY BLOCK-----
 EOF
 
+# write FIPS PPA files if the current build is a FIPS build
+if [[ ${SNAP_FIPS_BUILD+x} ]]; then
+    # for private builds a conf file is neccessary, setup for PPA access
+    # if provided
+    if [ -e etc/apt/auth.conf.d/01-fips.conf ]; then
```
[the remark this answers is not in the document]
valentindavid: And if we copy it from the makefile, the makefile should remove it.

static/etc/apt/auth.conf.d/01-fips.conf.example
valentindavid, line 1: Ah, I did not see, it is in static/. Please move this file in the root of the project. We should copy it in and remove it before packing.

snapcraft.yaml
Meulengracht: I'll add a line to remove it in that case - but no, I've been testing by doing `git remote set-url origin ...` to rename it and then just rename it back

Makefile
Meulengracht: Can we keep the `ifneq` ? It feels more explicit that this file is optional, and then we don't try to export a variable that might not exist

hooks/001-extra-packages.chroot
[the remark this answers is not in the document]
Meulengracht: added!

static/etc/apt/auth.conf.d/01-fips.conf.example
Meulengracht: I added a line stating the comments should be removed

Meulengracht: added, nice catch!

Makefile
valentindavid: What about `ifdef SNAP_FIPS_BUILD` to guard the export? That is easier to read than `ifneq (,$(wildcard ./.fips-env))`.

Makefile
```diff
@@ -43,6 +66,9 @@ install:
 	done
 	rm -rf $(DESTDIR)/install-data
 
+	# remove the auth file again
+	rm $(DESTDIR)/etc/apt/auth.conf.d/01-fips.conf || true
```
valentindavid, line 70: `rm -f` rather than `|| true`.

Makefile
Meulengracht: I'd say that is okay as long as we only export one variable, otherwise it quickly gets very verbose

valentindavid (approved):
Great!

hooks/001-extra-packages.chroot
```diff
@@ -187,7 +241,15 @@ case "$(dpkg --print-architecture)" in
         ;;
 esac
 
-apt-get install --no-install-recommends -y "${PACKAGES[@]}"
+if [[ ${SNAP_FIPS_BUILD+x} ]]; then
+    # Ensure vital crypt packages are refreshed / downgraded and downloaded
+    # from the FIPS ppa. This should also container openssh-server, but we already
```
alfonsosanchezbeato, line 246: typo container -> contain

alfonsosanchezbeato (commented):
looks good! Some minor things

alfonsosanchezbeato:
Couple of other comments I have:
* I think you should propose to main too, even if we will not use it there for the moment
* Now we use the commit description in the changelog, so we should try to have a cleaner history now or do squash merges

Makefile
```diff
@@ -33,6 +48,14 @@ install:
 		mknod -m 666 $(DESTDIR)/dev/urandom c 1 9
 	# copy static files verbatim
 	/bin/cp -a static/* $(DESTDIR)
+	# copy the FIPS PPA config file in if it exists and if
+	# the current build is a FIPS build
+ifdef SNAP_FIPS_BUILD
+	if [ -e ./fips.conf ]; then \
```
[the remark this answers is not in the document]
Meulengracht: The fips.conf is only for local testing

PR #255, opened 2024-09-27
static/run-snapd-from-snap: adapt to new mount place for snapd snap

alfonsosanchezbeato:
Now we will mount the snapd snap in /snap/snapd/<revision> on first boot to avoid duplicate mounts. Adapt to this initramfs changes.

### DIFF
--- a/static/usr/lib/core/run-snapd-from-snap
+++ b/static/usr/lib/core/run-snapd-from-snap
@@ -5,23 +5,32 @@
 
 set -eux
 
-# run_on_unseeded will mount/run snapd on an unseeded system
+# run_on_unseeded will run snapd on an unseeded system. The snapd snap
+# is expected to have been mounted and current link is expected to
+# have been created by the initramfs.
 run_on_unseeded() {
-    SNAPD_BASE_DIR="/run/mnt/snapd"
-
-    # We need to initialize /snap/snapd/current symlink so that the
-    # dynanic linker
-    # /snap/snapd/current/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
-    # is available to run snapd.
-    [ -d /snap/snapd ] || mkdir -p /snap/snapd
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
This would need to be backported to all core20+ bases. The change is compatible with older snap-bootstrap and with the new one with the changes in https://github.com/canonical/snapd/pull/14548.

static/usr/lib/core/run-snapd-from-snap
```diff
@@ -5,31 +5,40 @@
...
+    SNAPD_BASE_DIR="/snap/snapd/current"
+    if ! mountpoint -q "$SNAPD_BASE_DIR"; then
+        # Compatibility with old way where we had a duplicated mount in /run
+        # and "current" link was not created by initramfs.
+        SNAPD_BASE_DIR="/run/mnt/snapd"
+        # We need to initialize /snap/snapd/current symlink so that the
+        # dynamic linker
+        # /snap/snapd/current/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
+        # is available to run snapd.
+        [ -d /snap/snapd ] || mkdir -p /snap/snapd
```
valentindavid, line 21: Think about it, the condition was useless here, right? Maybe you can remove it in the same time.

valentindavid (approved):

Meulengracht (approved):
lgtm, thank you

static/usr/lib/core/run-snapd-from-snap
alfonsosanchezbeato: Good point, removed now.

alfonsosanchezbeato:
I've rebased this now and added some minor changes to be more flexible as we might have different listening socket in the future - asking for a re-review.

valentindavid (approved):
Looks fine. Just one question about the multiple sockets (you did not add it in this PR, but I am still wondering).

Meulengracht (approved):

PR #258, opened 2024-10-07
hooks: update fips repo details

Meulengracht:
We were using the wrong FIPS repository it was reported. Update details, unfortunately we do not have local access to this repo yet, but I've requested that.


Meulengracht:
Closing this, new feedback indicates we were using the right one

PR #259, opened 2024-10-09
static: create device nodes if nvidia driver is loaded

alfonsosanchezbeato:
Do the same actions as the 71-nvidia.rules files shipped by nvidia-kernel-common-* in case the nvidia driver has been loaded in the system.


Meulengracht (approved):
lgtm

static/usr/lib/udev/rules.d/71-nvidia.rules
```diff
@@ -1,11 +1,30 @@
+## This is an adapted version of 71-nvidia.rules file shipped by nvidia-kernel-common-*
```
valentindavid, line 1: I think it would be nice to say here what is adapted. Just to make it easier to update.

valentindavid: I see the path to nvidia-create-device is different. And  nvidia-persistenced is removed.

Why is nvidia-persistenced removed?

static/usr/bin/nvidia-create-device
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -eux
```
valentindavid, line 1: Why cannot we just ship ub-create-device?

static/usr/lib/udev/rules.d/71-nvidia.rules
valentindavid: I suppose for the persistenced,  there is no way for it to be shipped. But since it is is just a `ENV{SYSTEMD_WANTS}`, I am not sure this would cause any error.

alfonsosanchezbeato: I've commented on the things we change.

static/usr/bin/nvidia-create-device
alfonsosanchezbeato: The problem with it is that it depends on the nvidia driver version, so you get different version for 535, 550, etc. The real solution would be to let the kernel add udev rules to the system, but we are not there yet.

valentindavid: It is this right? https://github.com/canonical/nvidia-graphics-drivers/blob/560/debian/device-create/ub-device-create.c

That file has not changed for 4 years.

static/usr/lib/udev/rules.d/71-nvidia.rules
valentindavid: Thank you.

zyga (commented):
Is the order of calls to modprobe and Nvidia device create program relevant?

static/usr/bin/nvidia-create-device
alfonsosanchezbeato: Thanks, I actually thought this was coming from nvidia. I've changed this now.

valentindavid (approved):
Thank you for the changes.

alfonsosanchezbeato:
Fwiw, this has been tested on a device with nvidia GPU and nvidia modules.

PR #260, opened 2024-10-09
static: mount some directories from initramfs

alfonsosanchezbeato:
We want to mount some folders from the initramfs as they
are a requisite to mount the kernel snap from initramfs:
- /snap is the root of the mountpoints for the kernel snap and
  kernel-modules components
- /var/lib/snapd contains the drivers tree and the snap and component
  files
- /var/snap might contain dynamically generated kernel modules, and we
  want the links from the drivers tree to be valid since the moment we
  switch root.

Backport from https://github.com/canonical/core-base/pull/245


Meulengracht (approved):
thank you, +1

PR #261, opened 2024-10-14
Revert "Change netplan apply to workaround LP#2058976"

valentindavid:
This reverts commit 14ece5ff92726b5274e5fee07bd8a94fee3475a0.

This seems to be applied already.


alfonsosanchezbeato (changes requested):
I do not see the patch applied in version `0.106.1-7ubuntu0.22.04.4` of netplan.io, that is the one in jammy-updates. What was failing here? There is a new version in jammy-proposed (`0.107.1-3ubuntu0.22.04.1`) that might have the patch, are we pulling something from -proposed?

PR #264, opened 2024-10-16
run-snapd-from-snap: seed with access to keyring

valentindavid:
The new FDE manager needs access the kernel keyring to access the primary key. That includes seeding.

This is a backport of #229


alfonsosanchezbeato (approved):
LGTM

PR #266, opened 2024-10-17
static: create device nodes if nvidia driver is loaded

alfonsosanchezbeato:
Import the 71-nvidia.rules file shipped by nvidia-kernel-common-* and the ancillary ub-device-create executable that creates some device nodes when the nvidia kernel module is loaded.

Backport of https://github.com/canonical/core-base/pull/259


hooks/001-extra-packages.chroot
```diff
@@ -258,3 +260,11 @@ apt-get download snapd
 dpkg --fsys-tarfile snapd_*.deb |
        tar xf - ./usr/lib/systemd/system-generators/snapd-generator
 rm snapd_*.deb
+
+# Copy nvidia udev rules and dependencies (sources for this come from
+# Canonical, not from Nvidia, see
+# https://github.com/canonical/nvidia-graphics-drivers).
+nvidia_common_pkg=nvidia-kernel-common-550
```
Meulengracht, line 267: Is this something we need to keep maintained/updated or should 71-nvidia.rules stay relatively intact per LTS

alfonsosanchezbeato: It will have the usual updates from the deb if any. If you are thinking about a newer nvidia driver release, I do not think it happens too often.

Meulengracht (approved):

PR #267, opened 2024-10-22
core22: add FIPS build support

Meulengracht:
This is a backport of the core24 FIPS PR https://github.com/canonical/core-base/pull/249 .

The major difference here is that we pull the correct core-base to do changelog against for FIPS, and that we use a different PPA.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,18 @@
 # dir that contans the filesystem that must be checked
 TESTDIR ?= "prime/"
 SNAP_NAME=core22
+CODENAME:="$(shell . /etc/os-release; echo "$$VERSION_CODENAME")"
+SNAP_CORE_TRACK:=latest
+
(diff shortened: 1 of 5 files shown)


fips.conf
```diff
@@ -0,0 +1 @@
+machine private-ppa.launchpadcontent.net/ubuntu-advantage/pro-fips-updates login the-meulengracht password RVtd2H1Fvh4znwcZ7Nqn
```
zyga, line 1: Oops?

zyga (changes requested):

fips.conf
Meulengracht: !!!!

Meulengracht: Removed it, deleting your comment to remove it from history

valentindavid (approved):

PR #272, opened 2024-11-07
hooks: move pam.d files into /usr/lib/pam.d

Meulengracht:
Leave /etc/pam.d empty for Core Desktop to experiment with their own per-snap profiles.

A little info surrounding profile loading, also see the [man-pages](https://manpages.opensuse.org/Tumbleweed/pam-manpages/pam.d.5.en.html)

There are three directories, in which PAM is looking for the service files in the following order:

* /etc/pam.d
* /usr/lib/pam.d

If a file is not found in /etc/pam.d, it will be searched in /usr/lib/pam.d and afterwards in /usr/etc/pam.d. 

Considering this function in the source: https://github.com/linux-pam/linux-pam/blob/37b416c152ed833dcc55c84068c77eb713748d24/libpam/pam_handlers.c#L293 
it looks like this should be no issues

**UPDATE** It turned out that include files from pam.d files do only check in `/etc/pam.d`, which on the surface looks like to be a misalignment with the man-pages. Filed bug https://bugs.launchpad.net/ubuntu/+source/pam/+bug/2087827

REF: SNAPDENG-34125

### DIFF
--- a/hook-tests/031-faillock.test
+++ b/hook-tests/031-faillock.test
@@ -2,4 +2,4 @@
 
 set -eu
 
-grep pam_faillock.so etc/pam.d/common-auth
+grep pam_faillock.so usr/lib/pam.d/common-auth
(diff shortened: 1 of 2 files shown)


alfonsosanchezbeato (approved):
LGTM, thanks

PR #276, opened 2024-11-13
Remove the fips configuation check

alanmcanonical:
According to the building log of lp builder, the fips database is not updated during 001-extra-packages.chroot. So we try to remove the inside check for adding fips ppa into source list and **assume** that any lp building is granted access to current ppa internally

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -119,12 +119,8 @@ EOF
 
 # write FIPS PPA files if the current build is a FIPS build
 if [[ ${SNAP_FIPS_BUILD+x} ]]; then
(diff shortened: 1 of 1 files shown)


Meulengracht (approved):
Thank you for looking into the LP build issues. For context: The PPA is injected into the build container, however when we `chroot` for the hooks, the PPA is not propagated into the chroot, giving weird results.

So, we need to add the FIPS PPA in any case as long as it's a FIPS build, not just for the local one. The only difference between the local build now and the LP is whether we have the local fips.conf file for providing credentials. The LP builders should not need this.

valentindavid (approved):

PR #281, opened 2024-11-15
Makefile,snapcraft: fix the copy of buildd sources, do not setup FIPS PPA on LP

Meulengracht:


Makefile
```diff
@@ -28,9 +29,9 @@ install:
 	# ensure resolving works inside the chroot
 	cat /etc/resolv.conf > $(DESTDIR)/etc/resolv.conf
 	# copy-in launchpad's build archive
-	if grep -q ftpmaster.internal /etc/apt/sources.list; then \
+	if grep -q ftpmaster.internal /etc/apt/sources.list.d/lp-buildd.sources; then \
```
Meulengracht, line 32: From Noble (24.04) buildd no longer puts sources into `sources.list` but instead truncates this file and writes `/etc/apt/sources.list.d/lp-buildd.sources`

Makefile
```diff
@@ -28,9 +29,9 @@ install:
...
 		cp /etc/apt/sources.list $(DESTDIR)/etc/apt/sources.list; \
-		cp /etc/apt/trusted.gpg $(DESTDIR)/etc/apt/ || true; \
```
Meulengracht, line 33: This file no longer exists

Makefile
```diff
@@ -28,9 +29,9 @@ install:
...
+		cp /etc/apt/sources.list.d/lp-buildd.sources $(DESTDIR)/etc/apt/sources.list.d/lp-buildd.sources; \
```
bboozzoo, line 34: does this need to be conditional?

Makefile
```diff
@@ -10,6 +10,7 @@ CODENAME:="$(shell . /etc/os-release; echo "$$VERSION_CODENAME")"
 -include .fips-env
 ifdef SNAP_FIPS_BUILD
     export SNAP_FIPS_BUILD
+    export SNAP_BUILD_NAME
```
bboozzoo, line 13: why not always do the export? and if not, the comment right before `-include` needs to be updated

Makefile
Meulengracht: should probably match the existing pattern yes

bboozzoo (approved):

valentindavid (approved):

PR #282, opened 2024-11-15
FIPS packages: Replace existing packages with meta package

alanmcanonical:
### Description

- Add openssh to install the fips-variant ssh

### Rationale

- The normal openssh has issue to work properly with fips kernel snap


alanmcanonical:
The debian/control control the dependencies of fips-pkg and debian/substvars of ubuntu-fips controls the version of fips related package

Meulengracht (approved):

PR #284, opened 2024-11-18
hooks(core22): do not remove libapt for now

Meulengracht:
See https://github.com/canonical/core-base/issues/283 and the comment for the change.

I don't think its the right choice for the snapcraft snap to be importing this. On Chisel there will be no apt environment at all, and they cannot even assume any `apt` presence. I think we should file a bug for snapcraft to stop importing this, or atleast make them resistent to this. (See https://github.com/canonical/snapcraft/issues/5146)


alfonsosanchezbeato (approved):
LGTM, thanks

PR #285, opened 2024-11-18
hooks(core24): keep libapt around for now

Meulengracht:
See https://github.com/canonical/core-base/issues/283 and the comment for the change.

I don't think its the right choice for the snapcraft snap to be importing this. On Chisel there will be no apt environment at all, and they cannot even assume any apt presence. Filed a bug for Snapcraft: https://github.com/canonical/snapcraft/issues/5146

### DIFF
--- a/hooks/600-no-debian.chroot
+++ b/hooks/600-no-debian.chroot
@@ -6,8 +6,11 @@ set -ex
 
 echo "I: Removing the debian legacy"
 
-# dpkg-deb and dpkg purposefully left behind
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
Do we need this for core24? snapcraft is using base core22 atm, when they switch to core24 they should include libapt in their snap.

alfonsosanchezbeato (approved):
LGTM, after discussion agreed we need this back

PR #287, opened 2024-11-19
hooks: always write the FIPS preference for FIPS builds

Meulengracht:
While we fixed the FIPS build, and correctly can access the PPA - it seems I was missing this commit for our upstream version. This uses the correct pinning for FIPS builds.


bboozzoo (approved):
LGTM

valentindavid (approved):

PR #293, opened 2025-01-07
hooks: Add kcapi-tools to FIPS extra packages.

Ek0n:


Ek0n:
Apologies, I need to review if only a subset of the kcapi packages is needed.
I'll close the PR and reopen it after checking.

PR #296, opened 2025-01-09
console-conf: make exit status 22 not reported as a failure

valentindavid:
This is a quick fix that should be applied to upstream.

24 and main will also need a similar fix.


valentindavid:
> thank you, the fix.conf must be removed again once subiquity is updated?

the whole commit should be reverted.

alfonsosanchezbeato (approved):
LGTM - maybe for main we can wait for the subiquity fix and propose only for core24 too.

PR #298, opened 2025-01-13
Remove the Updates pocket to avoid dependency issue

alanmcanonical:
According to lp build log, the Updates pocket wants to introduce the newer version of systemd which causes the build failure. So this pr remove the Updates pocket


Meulengracht (approved):
Thank you for doing the change!

PR #307, opened 2025-01-27
hooks: copy the desired default font to Plymouth.ttf

alfonsosanchezbeato:
Previously we were moving, but now in noble Ubuntu-R.ttf was actually a link to Ubuntu[wdth,wght].ttf, so actually a symlink to a non-present file was created. Instead, copy the file that we have now, which will also fix the issue if in the future it becomes a symlink too.

Backport of https://github.com/canonical/core-base/pull/306


alfonsosanchezbeato:
The full fix for UC24 need the fix for https://bugs.launchpad.net/ubuntu/+source/plymouth/+bug/2096806 too.

Meulengracht (approved):

PR #308, opened 2025-01-28
hooks: remove sub directories and symlinks from /usr/share/doc

simondeziel:
Docs are mostly dead weight in the core artifact:

```
$ du -shc /snap/core24/current/usr/share/doc/
2.6M	/snap/core24/current/usr/share/doc/
2.6M	total
```


simondeziel:
Looking at the build failure, it seems the partial content under `/usr/share/doc` is intentionally there. As such, I'll close this PR.

PR #313, opened 2025-02-20
hooks,static: make /etc/vconsole.conf and /etc/locale.conf symlinks to writable

Meulengracht:
See the LP bug for more information, along with test results of this change: https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/2091657

The test results are in this comment: https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/2091657/comments/2

### DIFF
--- a/hooks/010-locale-vconsole-symlinks.chroot
+++ b/hooks/010-locale-vconsole-symlinks.chroot
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -eu
+
+# setup default console configuration, make them symlinks
+# into writable so they can be changed
+printf 'LANG=C.UTF-8\n' > /etc/writable/locale.conf
+ln -s ../writable/locale.conf /etc/default/locale
+ln -s writable/locale.conf /etc/locale.conf
+
+# setup default vconsole/keyboard configuration, make it a
+# symlink as well so it can change
+mv /etc/vconsole.conf /etc/writable/vconsole.conf
+ln -s ../writable/vconsole.conf /etc/default/keyboard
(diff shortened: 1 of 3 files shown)


Meulengracht:
This cannot be merged before the systemd patch has come upstream

hooks/010-locale-vconsole-symlinks.chroot
```diff
@@ -0,0 +1,15 @@
...
+# into writable so they can be changed
+printf 'LANG=C.UTF-8\n' > /etc/writable/locale.conf
+ln -s ../writable/locale.conf /etc/default/locale
+ln -s writable/locale.conf /etc/locale.conf
+
+# setup default vconsole/keyboard configuration, make it a
+# symlink as well so it can change
+mv /etc/vconsole.conf /etc/writable/vconsole.conf
+ln -s ../writable/vconsole.conf /etc/default/keyboard
+ln -s writable/vconsole.conf /etc/vconsole.conf
```
alfonsosanchezbeato, line 15: What is vconsole.conf here? In current core24 revisions it is a broken symlink to `default/keyboard`, will this fix that?

Meulengracht:
> Thanks, one question. Also, this should be proposed to main branch too. Afaiu, with this then we do not need the systemd patch mentioned in the bug anymore, is that correct?

We still need the systemd patch to be SRUd, because currently systemd does not support /etc/console.conf and /etc/vconsole.conf being symlinks.

alfonsosanchezbeato (approved):
Thanks for the responses - please remember to propose to main too.

valentindavid (approved):

PR #320, opened 2025-03-04
SECURITY.md: add our security policy

Meulengracht:
When approved, will be ported to `core24` and `main` branches

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,46 @@
+# Security policy
+
+## Supported versions
+<!-- Include start supported versions -->
+The release model of the core22 snap is following rolling releases. 
+The snap is released into the edge channel, and automatically promoted to 
+the beta channel if a snap revision is not already in progress for beta validation. 
+When reporting security issues against the core22 snap, only the latest 
+release of the core22 snap is supported.
+
+The core22 snap has regular releases that are fully automated. There are two 
+types of security fixes that can be shipped with new versions of the core22 snap.
+
+- Security fixes that are relevant to the files inside this repository.
+- Security fixes that are carried from the official archives. I.e security fixes 
+from debian packages carried inside the core22 snap.
+
+<!-- Include end supported versions -->
+
+## What qualifies as a security issue
+
+Security vulnerability that apply to packages in the Jammy archives also shipped by the
+core22 snap. Any vulnerability that allows the core22 snap to interfere outside 
+of the intended restrictions also qualifies as a security issue.
+
+## Reporting a vulnerability
+
+The easiest way to report a security issue is through
+[GitHub](https://github.com/canonical/core-base/security/advisories/new). See
+[Privately reporting a security
+vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+for instructions.
+
+Alternatively, please email [security@ubuntu.com](mailto:security@ubuntu.com) with a description of the issue, the
+steps you took to create the issue, affected versions, and, if known, mitigations for the issue.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,46 @@
+# Security policy
+
+## Supported versions
+<!-- Include start supported versions -->
+The release model of the core22 snap is following rolling releases. 
+The snap is released into the edge channel, and automatically promoted to 
+the beta channel if a snap revision is not already in progress for beta validation.
```
alfonsosanchezbeato, line 7: To be clear, there is an "edge validation" process that happens before promotion to beta, so some tests are performed to the edge snap too. Maybe we can also mention that additional tests involving real hw are performed in beta before promotion to candidate. And that we let the snap age in candidate for a while before promoting to stable, so candidate can get some external testing.

alfonsosanchezbeato (commented):
Thanks for this! Comment about promotion. Otherwise I would prefer to propose this sort of changes to main branch first to avoid forgetting to forward port, unless there is a specific reason to avoid that.

SECURITY.md
```diff
@@ -0,0 +1,46 @@
...
+- Security fixes that are carried from the official archives. I.e security fixes 
+from debian packages carried inside the core22 snap.
+
+<!-- Include end supported versions -->
+
+## What qualifies as a security issue
+
+Security vulnerability that apply to packages in the Jammy archives also shipped by the
+core22 snap. Any vulnerability that allows the core22 snap to interfere outside 
+of the intended restrictions also qualifies as a security issue.
```
alexmurray, line 24: Also any vulnerability that allows an unprivileged user on the local system to escalate privileges or cause a denial of service etc due to the use of the contents of the core22 snap on the system.

SECURITY.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Reporting a vulnerability
+
+The easiest way to report a security issue is through
+[GitHub](https://github.com/canonical/core-base/security/advisories/new). See
+[Privately reporting a security
+vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+for instructions.
+
```
alexmurray, line 33: I suggest removing the security@ubuntu.com part and sticking with github security advisories to coordinate communications.

alexmurray (changes requested):

SECURITY.md
Meulengracht: Good point, synced with Sergio to understand the edge-validation better as well

Meulengracht:
> Thanks for this! Comment about promotion. Otherwise I would prefer to propose this sort of changes to main branch first to avoid forgetting to forward port, unless there is a specific reason to avoid that.

Let me propose for main and core24 before this is merged once this is approved

alfonsosanchezbeato (approved):
LGTM, thanks

SECURITY.md
```diff
@@ -0,0 +1,50 @@
...
+- Security fixes that are relevant to the files inside this repository.
+- Security fixes that are carried from the official archives. I.e security fixes 
+from debian packages carried inside the core22 snap.
+
+<!-- Include end supported versions -->
+
+## What qualifies as a security issue
+
+Security vulnerability that apply to packages in the Jammy archives also shipped by the
```
alexmurray, line 27: s/vulnerability/vulnerabilities/

alexmurray, line 27: s/Jammy/Ubuntu 22.04 LTS/

alexmurray (changes requested):
Just a couple minor suggestions, otherwise LGTM

PR #324, opened 2025-03-13
core24: rollback pamd workaround and wait for SRU fix instead

Meulengracht:
This broke pam.d profiles that came with snaps since they load using relative paths. Relative paths do not work correctly right now and only expect profiles to exist in /etc/pam.d.


alfonsosanchezbeato (approved):
LGTM

PR #330, opened 2025-04-24
hooks: move around symlinks for the ssh.socket unit

alfonsosanchezbeato:
We want the symlinks for the ssh service in the systemd distro directory instead of in the admin directory. Previously, these links were copied to /etc/systemd on first boot, which is an issue as these symlinks are new to core24, so when remodeling from 20/22 to 24 they are not copied around as handle-writable-paths sees that the /etc/systemd folder already exists. This change ensures that the links are static and will always be present. This is not a problem when enabling/disabling ssh as for that the services rely on the presence of the /etc/ssh/sshd_not_to_be_run file, and in any case is a step forward in the path of having an empty /etc on installation in the future.

When upgrading core24 to one with this change this should be ok as we simply will have the same links in /etc/systemd/system/.

### DIFF
--- a/hooks/032-move-ssh-config.chroot
+++ b/hooks/032-move-ssh-config.chroot
@@ -1,0 +1,22 @@
+#!/bin/sh -ex
+
+# We want the symlinks for the ssh service in the systemd distro directory
+# instead of in the admin directory. Previously, these links were copied to
+# /etc/systemd on first boot, which is an issue as these symlinks are new to
+# core24, so when remodeling from 20/22 to 24 they are not copied around as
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
There are more issues though when remodeling from 20/22 to 24, like wrong motd message (old one is used).

valentindavid (approved):

Meulengracht (approved):

pedronis (approved):
look reasonable, I suppose this in itself is not an issue when upgrading the base on existing systems?

PR #332, opened 2025-04-25
many: put default motd in /usr

alfonsosanchezbeato:
Instead of having a symlink in /etc/motd that points to a writable path, put the default motd in /usr/lib/motd.d/50-default file, and add /etc/motd.d to the writable paths. With this

- When remodeling, we will show the message from the new base as it is part of the immutable directories.

- It is still possible to override the default message by creating a file named /etc/motd.d/50-default.

- It is also possible to add messages by creating additional files in /etc/motd.d/ or by adding scripts to /etc/update-motd.d/ as in the past.


hook-tests/014-set-motd.test
```diff
@@ -2,12 +2,15 @@
...
-test ! -e etc/update-motd.d/00-header
-test ! -e etc/update-motd.d/10-help-text
 test ! -e usr/lib/systemd/system/motd-news.service
 test ! -e usr/lib/systemd/system/motd-news.timer
 
-! grep -q unminimize etc/update-motd.d/*
+# Checks for empty update-motd.d - for future bases this will be a canary to
+# check if we need to leave anything new there.
+# shellcheck disable=SC2034 # read needs a variable
+if find etc/update-motd.d -mindepth 1 -maxdepth 1 | read -r var
```
valentindavid, line 14:
```
test -d etc/update-motd.d
test ! -s etc/update-motd.d
```

hooks/014-set-motd.chroot
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash -ex
 
-cat >/etc/motd<<EOF
+# Writing motd in this folder will allow overrides by creating the
+# /etc/motd.d/motd file.
+mkdir -p /usr/lib/motd.d
+cat > /usr/lib/motd.d/50-default <<EOF
```
valentindavid, line 6: Could this go in `factory/`?

valentindavid (approved):
Looks good. Few details...

hooks/014-set-motd.chroot
alfonsosanchezbeato: Good point, done

hook-tests/014-set-motd.test
alfonsosanchezbeato: I did a quick try but it looks like `-s` does not work for directories, or at least it give the same result for empty or non-emoty directory.

valentindavid: `/bin/sh` -> `/bin/bash` maybe? I would not be surprised dash misbehaves.

valentindavid: Ah no, you are right. It depends on the filesystem. It happens to work on btrfs.

Meulengracht (approved):

hook-tests/014-set-motd.test
alfonsosanchezbeato: Wow, that is weird... good to know

PR #335, opened 2025-04-30
hooks: write default configuration for vconsole.conf

Meulengracht:
The default shipped /etc/vconsole.conf is a symlink to default/keyboard, which is non-existant in the default setup. The default configuration is just to default to US keymap.

Add a test as my previous testing was obviously insufficient, and this fixes a weird symlink loop that was introduced earlier.

Port of https://github.com/canonical/core-base/pull/334 for main branch


alfonsosanchezbeato (approved):
LGTM, thanks

PR #336, opened 2025-05-01
hooks: make sure that the ssh.socket unit is enabled

alfonsosanchezbeato:
The recent change moving the symlinks enabling ssh.socket to /usr/lib made that
the unit was considered disabled as it was not considered static, although it
was still activated. This was an issue for snapd, that check is the unit is
enabled when changing ssh options. Fix by removing the [Install] section from
the unit.


alfonsosanchezbeato:
This needs backporting to core24 once approved.

PR #339, opened 2025-05-23
many: change name to core26

alfonsosanchezbeato:


spread.yaml
```diff
@@ -52,12 +52,12 @@ backends:
 
   lxd-nested:
     type: lxd
-    container-profiles: core24
+    container-profiles: core26
     systems:
       - ubuntu-24.04:
-          image: ucspread24
+          image: ucspread26
```
Meulengracht, line 58: this looks odd - the system is ubuntu-24.04? Maybe add a TODO that we need to add a new entry once an image exists?

alfonsosanchezbeato:
@Meulengracht thanks for the review, I've addressed your comments now

Meulengracht (approved):
Really nice - thank you!

PR #340, opened 2025-05-28
many: add tee-supplicant on arm64/armhf systems

alexclewontin:
This includes tee-supplicant as a service on arm64/armhf systems. tee-supplicant is primarily responsible for providing [Secure Storage](https://optee.readthedocs.io/en/latest/architecture/secure_storage.html) to OP-TEE Trusted Applications, and in the context of Ubuntu Core, specifically the https://github.com/OP-TEE/optee_ftpm TA.

It needs to be implemented as a service in the Core snap instead of as a snap app because, in the fTPM case, it needs to be available when snapd seals the encrypted partitions in install mode and our testing has shown that gadget services do not run early enough to guarantee availability. We also played with the concept of [systemd's Root Storage Daemons](https://systemd.io/ROOT_STORAGE_DAEMONS/) but unfortunately while we can cause the service to survive the initial switchroot killing spree, the second instance of systemd kills it on deserialization as the unit file is no longer there.

This PR adds about 24KB to the compressed size of the arm64 core snap (potentially less with the chiseling done in https://github.com/canonical/core-base/pull/225, I also have slice yaml files available for these packages!). It will not affect architectures other than arm64/armhf.

It is architected so as to not change the running behavior of systems that do not explicitly enable it by passing a data directory from the initrd mounted at /run/mnt/tee-data (see https://github.com/canonical/snapd/pull/15296 for corresponding changes to the core-initrd tooling).

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -244,6 +244,12 @@ case "$(dpkg --print-architecture)" in
         ;;
 esac
 
+case "$(dpkg --print-architecture)" in
+    arm64|armhf)
+        PACKAGES+=(tee-supplicant)
+        ;;
+esac
+
 if [[ ${SNAP_FIPS_BUILD+x} ]]; then
     # Ensure vital crypt packages are refreshed / downgraded and downloaded
     # from the FIPS ppa. This should also contain openssh-server, but we already
--- a/static/usr/lib/systemd/system/tee-supplicant.service.d/core-override.conf
+++ b/static/usr/lib/systemd/system/tee-supplicant.service.d/core-override.conf
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: BSD-2-Clause
+# Adapted from https://github.com/OP-TEE/optee_client/blob/6486773583b5983af8250a47cf07eca938e0e422/tee-supplicant/tee-supplicant%40.service.in
+[Unit]
+DefaultDependencies=no
+BindsTo=dev-teepriv0.device var-lib-optee\x2dclient-data-tee.mount
+After=dev-teepriv0.device var-lib-optee\x2dclient-data-tee.mount
+
+[Service]
+Type=notify
+ExecStartPre=-/usr/sbin/modprobe -v -r tpm_ftpm_tee
+ExecStartPost=-/usr/sbin/modprobe -v tpm_ftpm_tee
+ExecStop=-/bin/sh -c "/sbin/modprobe -v -r tpm_ftpm_tee ; /bin/kill $MAINPID"
--- a/static/usr/lib/udev/rules.d/99-optee.rules
+++ b/static/usr/lib/udev/rules.d/99-optee.rules
@@ -1,0 +1,5 @@
+# SPDX-License-Identifier: BSD-2-Clause
+# Adapted from https://github.com/OP-TEE/optee_client/blob/6486773583b5983af8250a47cf07eca938e0e422/tee-supplicant/optee-udev.rules.in
+KERNEL=="tee[0-9]*", MODE="0660", OWNER="root", GROUP="root", TAG+="systemd"
+KERNEL=="teepriv[0-9]*", MODE="0660", OWNER="root", GROUP="root", TAG+="systemd"
+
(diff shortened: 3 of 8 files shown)


static/usr/lib/systemd/system/tee-supplicant.service.d/core-override.conf
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: BSD-2-Clause
+# Adapted from https://github.com/OP-TEE/optee_client/blob/6486773583b5983af8250a47cf07eca938e0e422/tee-supplicant/tee-supplicant%40.service.in
+[Unit]
+DefaultDependencies=no
+BindsTo=dev-teepriv0.device var-lib-optee\x2dclient-data-tee.mount
```
valentindavid, line 5: Does it really BindsTo the mount rather than Requires?

Would `RequireMountsFor=/var/lib/optee-client/data/tee` not work?

static/usr/lib/systemd/system/tee-supplicant.service.d/core-override.conf
```diff
@@ -0,0 +1,12 @@
...
+After=dev-teepriv0.device var-lib-optee\x2dclient-data-tee.mount
+
+[Service]
+Type=notify
+ExecStartPre=-/usr/sbin/modprobe -v -r tpm_ftpm_tee
+ExecStartPost=-/usr/sbin/modprobe -v tpm_ftpm_tee
+ExecStop=-/bin/sh -c "/sbin/modprobe -v -r tpm_ftpm_tee ; /bin/kill $MAINPID"
```
valentindavid, line 12: `/usr/sbin/modprobe`?

static/usr/lib/systemd/system/tee-supplicant.service.d/core-override.conf
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: BSD-2-Clause
+# Adapted from https://github.com/OP-TEE/optee_client/blob/6486773583b5983af8250a47cf07eca938e0e422/tee-supplicant/tee-supplicant%40.service.in
+[Unit]
+DefaultDependencies=no
+BindsTo=dev-teepriv0.device var-lib-optee\x2dclient-data-tee.mount
+After=dev-teepriv0.device var-lib-optee\x2dclient-data-tee.mount
+
+[Service]
+Type=notify
```
valentindavid, line 9: I am not sure you should override this one. This is a drop-in. And the original looks like having that set.

static/usr/lib/systemd/system/tee-supplicant.service.d/core-override.conf
```diff
@@ -0,0 +1,12 @@
...
+ExecStartPre=-/usr/sbin/modprobe -v -r tpm_ftpm_tee
+ExecStartPost=-/usr/sbin/modprobe -v tpm_ftpm_tee
```
valentindavid, line 11: Not sure why we need that. So you want to remove `tpm_ftpm_tee` before starting. Then once `tee-supplicant` has started and sent `READY=1`, reload the module.

static/usr/lib/udev/rules.d/99-optee.rules
```diff
@@ -0,0 +1,5 @@
+# SPDX-License-Identifier: BSD-2-Clause
+# Adapted from https://github.com/OP-TEE/optee_client/blob/6486773583b5983af8250a47cf07eca938e0e422/tee-supplicant/optee-udev.rules.in
+KERNEL=="tee[0-9]*", MODE="0660", OWNER="root", GROUP="root", TAG+="systemd"
+KERNEL=="teepriv[0-9]*", MODE="0660", OWNER="root", GROUP="root", TAG+="systemd"
```
valentindavid, line 4: If group is root, maybe 0600 makes more sense.

hooks/001-extra-packages.chroot
```diff
@@ -244,6 +244,12 @@ case "$(dpkg --print-architecture)" in
         ;;
 esac
 
+case "$(dpkg --print-architecture)" in
+    arm64|armhf)
+        PACKAGES+=(tee-supplicant)
```
valentindavid, line 249: Maybe one day, we should make a list of all packages that are from universe rather than main.

static/usr/lib/systemd/system/tee-supplicant.service.d/core-override.conf
valentindavid: Eventually, we should maybe build it ourselves. That is a universe package anyway.

PR #344, opened 2025-06-09
.github/workflows/release.yaml: add release job

alfonsosanchezbeato:
For the moment it will be started manually.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,14 @@
+name: Release snap
+
+on: workflow_dispatch
+
+jobs:
+  build_release:
+    runs-on: ubuntu-latest
+    environment: store
+    steps:
+      - name: Build and release to beta channel
+        env:
+          LP_CREDENTIALS: ${{ secrets.LP_CREDENTIALS }}
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_CREDENTIALS }}
+        uses: snapcore/system-snaps-cicd-tools/action-rebuild-base@main
--- a/get-version.sh
+++ b/get-version.sh
@@ -1,0 +1,10 @@
+#!/bin/bash -ex
+
+tag=$(git tag --points-at HEAD | grep -E '^[0-9]{8}(-[0-9]+)?$' |
+          sort --field-separator=- --key=1n,1 --key=2n,2 | tail -n1 || printf "")
+if [ -z "$tag" ]; then
+    # Current date is the default if there are no date tags
+    tag=$(/bin/date +%Y%m%d)
+fi
+
+printf "%s" "$tag"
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,14 @@
+name: Release snap
+
+on: workflow_dispatch
+
+jobs:
+  build_release:
+    runs-on: ubuntu-latest
+    environment: store
+    steps:
+      - name: Build andr release to beta channel
```
valentindavid, line 10: typo?

get-tag.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# tags are returned in lexicographical order, so dates are ordered but sequence
+# numbers are not (that is, we will have 20250101-11 before 20250101-2), so we
+# need to look for the max seq number for the last date in the list.
+tags=$(git tag --points-at HEAD || printf "")
+# Current date is the default if there are no date tags
+date_last=$(/bin/date +%Y%m%d)
+last_tag="$date_last"
+seq_last=0
+while read -r tag; do
```
valentindavid, line 11: Could we just use `sort -t- -k 1n,1 -k 2n,2`?

.github/workflows/release.yaml
alfonsosanchezbeato: Fixed

get-version.sh
```diff
@@ -0,0 +1,21 @@
...
+git_date_tag() {
+    tags=$(git tag --points-at HEAD || printf "")
+    # We expect tags of format <date>-<sequence>. The sort commands orders them
+    # first by date and then by sequence.
+    while read -r tag; do
+        if [[ ! "$tag" =~ ^[0-9]{8}(-[0-9]+)?$ ]]; then
+            continue
+        fi
+        printf "%s\n" "$tag"
+    done <<< "$tags" | sort --field-separator=- --key=1n,1 --key=2n,2 | tail -n1
```
valentindavid, line 12: Why the loop? Cannot we filter with grep?

valentindavid: That said, to parse and sort, I would prefer a pascal program that a shell script.

Meulengracht (approved):
Nice with the changes, looks much clearer now :)

get-version.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -ex
+
+tag=$(git tag --points-at HEAD | grep -E '^[0-9]{8}(-[0-9]+)?$' |
+          sort --field-separator=- --key=1n,1 --key=2n,2 | tail -n1) || printf ""
```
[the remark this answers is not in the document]
valentindavid: But that does not go in tag. You wrote something like

```
tag=$(command) || printf ""
```

That means `tag` would never get what `printf` wrote. I think you maybe wanted to write: 
```
tag=$(command || printf "")
```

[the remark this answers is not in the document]
alfonsosanchezbeato: ok, I see what you mean now, I got confused with shell syntax here - fixed now, thanks

PR #353, opened 2025-07-01
static/secureboot-db.service: check mode by looking at modeenv

alfonsosanchezbeato:
Checking the kernel command line to know the mode is not enough as we could be installing UC from the initramfs. Look instead at the mode set in modeenv. We do it in ExecStart so the service does not fail - we cannot use ExecCondition as the upstream service sets SuccessExitStatus=1, so even on a failed condition the ExecStart command runs.

Checking the kernel command line was causing issues as when installing from the initramfs, the dbx was updated after keys were sealed, and the system was not able to unseal after a reboot as dbx was not the expected one.

Eventually we should move the dbx update to make it happen in the initramfs.


static/usr/lib/systemd/system/secureboot-db.service.d/core-override.conf
```diff
@@ -1,2 +1,3 @@
-[Unit]
-ConditionKernelCommandLine=snapd_recovery_mode=install
+[Service]
+ExecStart=
+ExecStart=sh -c 'if [ "$(/usr/libexec/core/get-mode mode)" = install ]; then /usr/bin/sbkeysync --no-default-keystores --keystore /usr/share/secureboot/updates --verbose; fi'
```
valentindavid, line 3: Why did not you use `ExecCondition` here like other places?

valentindavid:
I wonder in the long term we should look into enabling those service through a generator.

static/usr/lib/systemd/system/secureboot-db.service.d/core-override.conf
alfonsosanchezbeato: Right, that is a better option. Changed now, thanks.

PR #361, opened 2025-07-30
get-version.sh: filter by _$branch suffix when looking at tags

alfonsosanchezbeato:
We have a branch per base in the core-base repository, so we need to differentiate between date tags to avoid repetition. For this, we will add a _$branch suffix to tags.

Here we take that into account when filtering tags in the get-version.sh script.

### DIFF
--- a/get-version.sh
+++ b/get-version.sh
@@ -1,7 +1,12 @@
-#!/bin/bash -ex
+#!/bin/bash -exu
 
-tag=$(git tag --points-at HEAD | grep -E '^[0-9]{8}(-[0-9]+)?$' |
-          sort --field-separator=- --key=1n,1 --key=2n,2 | tail -n1 || printf "")
+# Limit what branch can be to avoid weird regex in the call to sed
+branch=$(git rev-parse --abbrev-ref HEAD | grep -E '^[a-z0-9/-]+$' || printf "")
+tag=
+if [ -n "$branch" ]; then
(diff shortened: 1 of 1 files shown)


get-version.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash -ex
 
-tag=$(git tag --points-at HEAD | grep -E '^[0-9]{8}(-[0-9]+)?$' |
+branch=$(git rev-parse --abbrev-ref HEAD)
```
valentindavid, line 3: Can we validate that? To make sure that it does not do any weird regex and we do not understand why it is doing something unexpected. Like limiting to `[a-z0-9/]+`.

get-version.sh
```diff
@@ -1,7 +1,12 @@
-#!/bin/bash -ex
+#!/bin/bash -exu
 
-tag=$(git tag --points-at HEAD | grep -E '^[0-9]{8}(-[0-9]+)?$' |
-          sort --field-separator=- --key=1n,1 --key=2n,2 | tail -n1 || printf "")
+# Limit what branch can be to avoid weird regex in the call to sed
+branch=$(git rev-parse --abbrev-ref HEAD | grep -E '^[a-z0-9/-]+$' || printf "")
```
Meulengracht, line 4: what is the trailing || printf "" doing?

valentindavid: Same as `|| true`

valentindavid (approved):

get-version.sh
alfonsosanchezbeato: Yes, you can do in different ways, it is just a way to ensure an empty branch variable if the grep command fails.

PR #365, opened 2025-08-25
static: copy udev disk rules from core-initrd

valentindavid:
Without matching rules in core-base, the device symlinks created in initrd creates device units which core-base's systemd does not what to do with, and are shown as "activating".


static/usr/lib/udev/rules.d/90-ubuntu-core-partitions.rules
alfonsosanchezbeato, line 1: Probably this should have a comment saying that it is a copy of the file in the initramfs and add a link to the file in the snapd repo, as it should be kept in sync.

alfonsosanchezbeato (commented):
Thanks, minor comment

PR #367, opened 2025-08-27
spread.yaml: Sync google-nested-arm with snapd

valentindavid:


Meulengracht (approved):
Thank you! The build is running now i can see. Let us see the tests what happens

PR #396, opened 2026-01-28
hooks: add script to remove unneeded apparmor profiles

zyga:
This speeds up boot time as we prevent many calls to apparmor_parser.

This patch is amended compared to the core26 counterpart, to provide the correct set of expected profiles (in this case, none). Those differ from base to base.


alfonsosanchezbeato (approved):
Thanks

Meulengracht:
@zyga there are some comments from copilot, some of them might be worth addressing - can you take a look then I'll review and merge

PR #418, opened 2026-03-27
static/usr/lib/udev/rules.d: replace partition scanning with symlink …

alfonsosanchezbeato:
…validation

Prevent running twice 'snap-boostrap scan', once in initramfs and next time from the rootfs. Instead, recreate the device units by using the symlinks already in /dev/disk. This ensures that the units are the same, is faster, and also prevents the issue that on first boot /usr/lib/snapd/snap-bootstrap is not yet available when the rules are run, as usr-lib-snapd.mount has not been created yet. This provoked error like:

zram0: Failed to find and pin callout binary "/usr/lib/snapd/snap-bootstrap": No such file or directory

### DIFF
--- a/static/usr/lib/udev/rules.d/90-ubuntu-core-partitions.rules
+++ b/static/usr/lib/udev/rules.d/90-ubuntu-core-partitions.rules
@@ -1,21 +1,21 @@
-# This file should be the same as in snapd/core-initrd
-
+# This file ensures that we keep the device units created by the file with the
+# same name in snapd/core-initrd.
 SUBSYSTEM!="block", GOTO="ubuntu_core_partitions_end"
 KERNEL=="loop*|mmcblk[0-9]boot[0-9]", GOTO="ubuntu_core_partitions_end"
 
-ENV{DEVTYPE}=="disk", IMPORT{program}="/usr/lib/snapd/snap-bootstrap scan-disk"
-ENV{DEVTYPE}=="partition", IMPORT{parent}="UBUNTU_DISK"
-ENV{UBUNTU_DISK}!="1", GOTO="ubuntu_core_partitions_end"
-
-ENV{DEVTYPE}=="disk", SYMLINK+="disk/snapd/disk"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-seed", SYMLINK+="disk/snapd/ubuntu-seed"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-boot", SYMLINK+="disk/snapd/ubuntu-boot"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-data", ENV{ID_FS_TYPE}=="crypto_LUKS", SYMLINK+="disk/snapd/ubuntu-data-luks"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-data", ENV{ID_FS_TYPE}!="crypto_LUKS", SYMLINK+="disk/snapd/ubuntu-data"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-save", ENV{ID_FS_TYPE}=="crypto_LUKS", SYMLINK+="disk/snapd/ubuntu-save-luks"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-save", ENV{ID_FS_TYPE}!="crypto_LUKS", SYMLINK+="disk/snapd/ubuntu-save"
+PROGRAM=="/usr/bin/sh -c '[ -L /dev/disk/snapd/disk ] && [ $$(readlink -f /dev/disk/snapd/disk) = /dev/%k ]'", \
+    SYMLINK+="disk/snapd/disk"
+PROGRAM=="/usr/bin/sh -c '[ -L /dev/disk/snapd/ubuntu-seed ] && [ $$(readlink -f /dev/disk/snapd/ubuntu-seed) = /dev/%k ]'", \
+    SYMLINK+="disk/snapd/ubuntu-seed"
+PROGRAM=="/usr/bin/sh -c '[ -L /dev/disk/snapd/ubuntu-boot ] && [ $$(readlink -f /dev/disk/snapd/ubuntu-boot) = /dev/%k ]'", \
+    SYMLINK+="disk/snapd/ubuntu-boot"
+PROGRAM=="/usr/bin/sh -c '[ -L /dev/disk/snapd/ubuntu-data-luks ] && [ $$(readlink -f /dev/disk/snapd/ubuntu-data-luks) = /dev/%k ]'", \
+    SYMLINK+="disk/snapd/ubuntu-data-luks"
+PROGRAM=="/usr/bin/sh -c '[ -L /dev/disk/snapd/ubuntu-data ] && [ $$(readlink -f /dev/disk/snapd/ubuntu-data) = /dev/%k ]'", \
+    SYMLINK+="disk/snapd/ubuntu-data"
+PROGRAM=="/usr/bin/sh -c '[ -L /dev/disk/snapd/ubuntu-save-luks ] && [ $$(readlink -f /dev/disk/snapd/ubuntu-save-luks) = /dev/%k ]'", \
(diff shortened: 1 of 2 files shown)


Meulengracht (approved):

static/usr/lib/udev/rules.d/90-ubuntu-core-partitions.rules
```diff
@@ -1,21 +1,21 @@
...
-ENV{UBUNTU_DISK}!="1", GOTO="ubuntu_core_partitions_end"
-
-ENV{DEVTYPE}=="disk", SYMLINK+="disk/snapd/disk"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-seed", SYMLINK+="disk/snapd/ubuntu-seed"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-boot", SYMLINK+="disk/snapd/ubuntu-boot"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-data", ENV{ID_FS_TYPE}=="crypto_LUKS", SYMLINK+="disk/snapd/ubuntu-data-luks"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-data", ENV{ID_FS_TYPE}!="crypto_LUKS", SYMLINK+="disk/snapd/ubuntu-data"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-save", ENV{ID_FS_TYPE}=="crypto_LUKS", SYMLINK+="disk/snapd/ubuntu-save-luks"
-ENV{DEVTYPE}=="partition", ENV{ID_PART_ENTRY_NAME}=="ubuntu-save", ENV{ID_FS_TYPE}!="crypto_LUKS", SYMLINK+="disk/snapd/ubuntu-save"
+PROGRAM=="/usr/bin/sh -c '[ -L /dev/disk/snapd/disk ] && [ $$(readlink -f /dev/disk/snapd/disk) = /dev/%k ]'", \
```
valentindavid: I would prefer we just skip (GOTO the end) if we are missing snap-boostrap. And retrigger once it appears.

Let me think about it. Could you split the PR in the mean time?

valentindavid: What is the case where we do not have snap-bootstrap? Is that when we are not seeded? It feels to me that would should mount the seed's snap in /usr/lib/snapd in initrd in that case.

valentindavid: No I am thinking... we need /snap/snapd/current symink for the binary to work... Maybe we should have snap-bootstrap (along with snap-repair in my opinion) in core-base.

valentindavid: Another alternative is to add persist tag.

valentindavid: Could you add `OPTIONS+="db_persist"` to those devices (after `ENV{UBUNTU_DISK}!="1", GOTO="ubuntu_core_partitions_end"`)

That should go in the rules of core-initrd.

alfonsosanchezbeato: snap-bootstrap is not available early when we have not seeded yet (snapd snap not mounted).

But still, now I actually prefer to not call it at all from here. These rules are just to ensure that the units created by the initramfs are not in a weird state ("activating..."). And for that, it is safer to look at the symlinks and reproduce the units as this does - no extra time spent calling snap-bootstrap plus it is ensured that we "detect" the same devices.

Other options that have these properties would be good too if they are possible.

alfonsosanchezbeato: if using db_persist, then this rules files could go away from the base?

alfonsosanchezbeato: Adding `OPTIONS+="db_persist"` to the rules in initramfs and removing the rules from the base, there are no "activating" units anymore, but then there is actually no `/dev/disk/snapd` folder after switching root, Which might be just ok, but maybe we could want the symlinks in the future to use that information from elsewhere.

alfonsosanchezbeato:
Fwiw, the problem with the units is not only the traces, but that the related units are shown as "activating":

```
● dev-disk-by\x2ddesignator-esp.device                                                loaded activating tentative /dev/disk/by-designator/esp
● dev-disk-snapd-disk.device                                                          loaded activating tentative /dev/disk/snapd/disk
● dev-disk-snapd-ubuntu\x2dboot.device                                                loaded activating tentative /dev/disk/snapd/ubuntu-boot
● dev-disk-snapd-ubuntu\x2ddata\x2dluks.device                                        loaded activating tentative /dev/disk/snapd/ubuntu-data-luks
● dev-disk-snapd-ubuntu\x2dsave\x2dluks.device                                        loaded activating tentative /dev/disk/snapd/ubuntu-save-luks
● dev-disk-snapd-ubuntu\x2dseed.device                                                loaded activating tentative /dev/disk/snapd/ubuntu-seed
```

alfonsosanchezbeato:
@valentindavid I've now changed the rules as discussed, although there were some caveats:
* `IMPORT{db}="UBUNTU_DISK"` was needed to get UBUNTU_DISK from the previous event
* `ATTR{ubuntu-disk}="$env{UBUNTU_DISK}"` does not work as the attribute needs to exist in sysfs
* I added a marker file as otherwise snap-bootstrap was still tried to run for not-boot disks, showing error traces (checking for UBUNTU_DISK just prevents re-running for that disk, so the "activating" issue do not happen even without the marker file)

static/usr/lib/udev/rules.d/90-ubuntu-core-partitions.rules
```diff
@@ -3,8 +3,20 @@
 SUBSYSTEM!="block", GOTO="ubuntu_core_partitions_end"
 KERNEL=="loop*|mmcblk[0-9]boot[0-9]", GOTO="ubuntu_core_partitions_end"
 
-ENV{DEVTYPE}=="disk", IMPORT{program}="/usr/lib/snapd/snap-bootstrap scan-disk"
+# Pull from db from a previous event for this device if it exists
+IMPORT{db}="UBUNTU_DISK"
 ENV{DEVTYPE}=="partition", IMPORT{parent}="UBUNTU_DISK"
+
+TEST=="/run/udev/snapd-scanned", GOTO="ubuntu_core_partitions_already_scanned"
```
valentindavid: Maybe `/run/snapd/ubuntu-disk-already-found` ?

static/usr/lib/udev/rules.d/90-ubuntu-core-partitions.rules
```diff
@@ -3,8 +3,20 @@
...
+
+ENV{DEVTYPE}=="disk", IMPORT{program}="/usr/lib/snapd/snap-bootstrap scan-disk"
+ENV{UBUNTU_DISK}=="1", RUN+="/usr/bin/touch /run/udev/snapd-scanned"
+
+LABEL="ubuntu_core_partitions_already_scanned"
+
+# Ensure UBUNTU_DISK survives udevadm info --cleanup-db
+OPTIONS+="db_persist"
```
valentindavid: This could go before the label. And since we use tag file, we could also just persist only if `ENV{UBUNTU_DISK}=="1"`.

static/usr/lib/udev/rules.d/90-ubuntu-core-partitions.rules
valentindavid: With a mkdir -p

valentindavid: I think we should also make snap-bootstrap create that tag file.

alfonsosanchezbeato: Done now

static/usr/lib/udev/rules.d/90-ubuntu-core-partitions.rules
alfonsosanchezbeato: It cannot be moved before the label as udevadm retrigger creates new events, so we need to ensure that UBUNTU_DISK is persisted whenever it was found. I've put it after `ENV{UBUNTU_DISK}!="1", GOTO="ubuntu_core_partitions_end"` though so persist happens only for events that have UBUNTU_DISK set.

valentindavid (approved):
Remember to also apply it to core-initrd in snapd.

PR #428, opened 2026-04-09
tools: do not use python3.12 feature since this builds with python3.10

andrewphelpsj:
Usage of nested quotes of the same type in format strings was introduced in https://peps.python.org/pep-0701. This isn't available until python3.12.

All core22 builds are failing because of this right now: https://code.launchpad.net/~ubuntu-core-service/+snap/core22


alfonsosanchezbeato (approved):
LGTM

Meulengracht (approved):

PR #430, opened 2026-04-15
static/usr/lib/wsl: create empty wsl dir in base

jpm-canonical:
See comment 1: https://github.com/canonical/snapd/pull/16726#discussion_r2964300217
See comment 2: https://github.com/canonical/snapd/pull/16726#issuecomment-4175511015

This PR adds the `/usr/lib/wsl` directory to the static content of the base. The directory is empty, except for a readme explaining what it is for. The content of `/usr/lib/wsl` from a WSL2 host will be mounted over this directory at runtime.

Please also cherrypick this commit on to core22, core24 and core26.

### DIFF
--- a/hooks/020-extra-files.chroot
+++ b/hooks/020-extra-files.chroot
@@ -55,3 +55,6 @@ chmod 0644 /usr/lib/systemd/system.conf.d/11-snapd-ctrl-alt-del-burst.conf
 
 mkdir -p /etc/iproute2
 mkdir -p /var/lib/dhcp
+
+echo "creating wsl mount point"
+mkdir -p /usr/lib/wsl
--- a/static/usr/lib/wsl/README
+++ b/static/usr/lib/wsl/README
@@ -1,0 +1,1 @@
+This directory is a placeholder for mounting runtime libraries from a WSL host.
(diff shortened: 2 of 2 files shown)


alfonsosanchezbeato (commented):
Please make the directory from 020-extra-files.chroot instead

alfonsosanchezbeato (approved):
LGTM, thanks

alfonsosanchezbeato:
@jpm-canonical thinking about this now, where does the content that you need to put it `/usr/lib/wsl` come from? Because if it comes from the distro, we could always include in directly in the base snap.

jpm-canonical:
Hi @alfonsosanchezbeato, the content comes from the WSL2 container, which on its turn gets it from the Windows on the host. So basically what we have is:

`[Windows GPU drivers] -> [WSL2 container] -> [snap]`

This needs to be done at runtime, as the libraries and drivers that are mounted in, depends very much on the Windows on the host system. For example, the libraries in `/usr/lib/wsl` are totally different between a host with NVIDIA GPU versus a host with Intel GPU. These libraries are also very dependent on which version of the GPU drivers are installed on the Windows host.

PR #432, opened 2026-04-16
snapcraft, hooks: include /usr/share/X11/XErrorDB

bboozzoo:
Include /usr/share/X11/XErrorDB (~42KB). This avoids needing to ship the
file in the GPU drivers snap (typically mesa-2604) and setting it up in
place with the help of snap layouts, which in turn would require
construction of a writable mimic on top of /usr/share/X11.

Related: https://github.com/canonical/snapcraft/pull/6188

### DIFF
--- a/hooks/602-cleanup-xkb.chroot
+++ b/hooks/602-cleanup-xkb.chroot
@@ -10,4 +10,3 @@ echo "I: Removing xkb data"
 # Remove /usr/share/xkeyboard-config-2 and the symlink to it (/usr/share/X11/xkb)
 rm -rv /usr/share/xkeyboard-config-2
 rm -v /usr/share/X11/xkb
-rmdir /usr/share/X11
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -138,6 +138,7 @@ parts:
         libpam-pwquality_libs
         libpciaccess0_libs
         libwrap0_host-files
+        libx11-data_xerrordb
         locales_standard
 
         netcat-openbsd_bins


alfonsosanchezbeato (commented):
Please make the directory from 020-extra-files.chroot instead

bboozzoo:
> Please make the directory from 020-extra-files.chroot instead

done

hooks/020-extra-files.chroot
```diff
@@ -55,3 +55,10 @@ chmod 0644 /usr/lib/systemd/system.conf.d/11-snapd-ctrl-alt-del-burst.conf
 
 mkdir -p /etc/iproute2
 mkdir -p /var/lib/dhcp
+
+echo "creating XErrorDB mount point"
+mkdir -p /usr/share/X11
+cat <<'EOF' > /usr/share/X11/XErrorDB
+! This is a placeholder file for bind mounting XErrorDB from GPU drivers
+! provider snap.
+EOF
```
alfonsosanchezbeato: Not sure if we need the file, on one side is nice but on the other hand we have plenty of mount points and we do not put files there. I think it is enough to have this as a comment in this script instead.

bboozzoo: The file is the mount point, https://github.com/canonical/snapcraft/pull/6188 which uses `bind-file` with `/usr/share/X11/XErrorDB` target

alfonsosanchezbeato: Got it, thanks for clarifying

alfonsosanchezbeato (dismissed):
LGTM, thanks

bboozzoo:
Per discussion with @Meulengracht we'll use chisel and ship the actual file

Meulengracht (approved):
Thank you for this! Appreciate the change on direction

alfonsosanchezbeato (approved):
LGTM, thanks, this is indeed better

PR #437, opened 2026-04-23
snapcraft.yaml: comment out isa-riscv64-rva23

alfonsosanchezbeato:
Comment out isa-riscv64-rva23 assumes. Our github build process will uncomment it only for riscv64 architecture. This is needed because snapd versions previous to 2.75 fail to install core26 if the flag is present, for any architecture. And in many cases the snapd present is the one from the debian package.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -9,7 +9,8 @@ build-base: core26
 grade: stable
 assumes:
   - snapd2.55.5
-  - isa-riscv64-rva23
+  # This is uncommented in our github builds for riscv64
+  ## - isa-riscv64-rva23
 
 parts:
   # TODO: rewrite this once all slices are upstreamed to use chisel slices


snapcraft.yaml
```diff
@@ -9,7 +9,8 @@ build-base: core26
 grade: stable
 assumes:
   - snapd2.55.5
-  - isa-riscv64-rva23
+  # This is uncommented in our github builds for riscv64
```
valentindavid, line 12: Where?

alfonsosanchezbeato: by modifying recipe in https://github.com/canonical/system-snaps-cicd-tools/blob/main/workflows/trigger-lp-build.py

alfonsosanchezbeato: (crating a new branch and a new recipe)

valentindavid (approved):

Meulengracht (approved):

PR #441, opened 2026-05-01
static: update default MOTD with Gemma4 AI example

alfonsosanchezbeato:
Replace IoT web-kiosk example with local AI box setup using Gemma4. Update installation instructions and add network access guidance with VM-specific documentation link.


valentindavid (approved):

Meulengracht (approved):
Haha, this is a more fun example and definitely more tinker-friendly :-)

PR #442, opened 2026-05-04
many: convert systemd generators from shell to rust

alfonsosanchezbeato:
This improves boot speed in resolute, as some commands usually used in shell scripts (mkdir, cat) have become slower there.


generators/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    let argv0 = args.first().map(String::as_str).unwrap_or("");
+    let generator_name = Path::new(argv0)
+        .file_name()
+        .and_then(|n| n.to_str())
+        .unwrap_or(argv0);
+
+    let normal_dir = args.get(1).map(String::as_str).unwrap_or("");
+    let early_dir = args.get(2).map(String::as_str).unwrap_or("");
+    let late_dir = args.get(3).map(String::as_str).unwrap_or("");
```
valentindavid: I think it should fail instead of using `""`. Sounds a bit dangerous to continue with "" as a path.

generators/src/generators/rtc_sys_time_init.rs
```diff
@@ -0,0 +1,124 @@
...
+        }
+        first = false;
+    }
+
+    result
+}
+
+fn log_kmsg(msg: &str) {
+    // Best-effort: write to kernel message buffer so the message survives
+    // before the journal is available.
```
valentindavid: I do not get this comment. Man page for `systemd.generator` says:

> Since `syslog(3)` is not available (see above), log messages have to be written to `/dev/kmsg` instead.

generators/src/kernel_args.rs
```diff
@@ -0,0 +1,84 @@
...
+    for param in parse_cmdline(&cmdline) {
+        let name_part = param.split('=').next().unwrap_or(&param);
+        if name_part.replace('_', "-") == looking_for {
+            let value = param
+                .find('=')
+                .map(|i| param[i + 1..].to_string())
+                .unwrap_or_default();
+            return Some(value);
+        }
+    }
```
valentindavid, line 30: I really doubt this is correct.

valentindavid: Ah sorry. Misread.

generators/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+    let result = match generator_name {
+        "bootchart" => generators::bootchart::run(normal_dir, early_dir, late_dir),
+        "bootloaders" => generators::bootloaders::run(normal_dir, early_dir, late_dir),
+        "hybrid-users" => generators::hybrid_users::run(normal_dir, early_dir, late_dir),
+        "console-conf-generator" => generators::console_conf::run(normal_dir, early_dir, late_dir),
+        "rtc-sys-time-init-generator" => {
+            generators::rtc_sys_time_init::run(normal_dir, early_dir, late_dir)
+        }
+        _ => {
+            eprintln!("Unknown generator: {}", generator_name);
```
valentindavid: stderr is null in a generator

generators/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+            process::exit(1);
+        }
+    };
+
+    if let Err(e) = result {
+        eprintln!("Generator {} failed: {}", generator_name, e);
```
valentindavid: Same here

generators/src/generators/rtc_sys_time_init.rs
alfonsosanchezbeato: Right, changed

generators/src/main.rs
alfonsosanchezbeato: Changed

generators/src/main.rs
alfonsosanchezbeato: Changed to using kmsg.

generators/src/main.rs
alfonsosanchezbeato: Changed to using kmsg.

alfonsosanchezbeato (commented):
Comments addressed now.

generators/src/util.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// Create a symlink at `dst` pointing to `src`, replacing any existing file
+/// or symlink at `dst`.
+pub fn symlink_force(src: &str, dst: &str) -> io::Result<()> {
+    match unix_fs::symlink(src, dst) {
+        Ok(()) => Ok(()),
+        Err(e) if e.kind() == io::ErrorKind::AlreadyExists => {
+            fs::remove_file(dst)?;
+            unix_fs::symlink(src, dst)
+        }
```
valentindavid: Could we just create a temporary symlink and rename it so that we do not have to retry and avoid TOCTOU bugs?

valentindavid: And of course always do that since the first attempt.

alfonsosanchezbeato: Excellent point, changed now

valentindavid (approved):

PR #453, opened 2026-07-06
snapcraft: adjust slices after 26-04 update

Meulengracht:
Our chisel fork has been rebased to incorporate the latest 26.04 changes. There has been some smaller reorginization of slices and thus some smaller changes have been made to build.

I also ended up switching the tests to openstack with this PR to make sure things were still building and working. The google backend can no longer be used for CI.

### DIFF
--- a/tests/lib/prepare-utils.sh
+++ b/tests/lib/prepare-utils.sh
@@ -158,6 +158,43 @@ get_core_snap_name() {
     echo "core26_${date}_$(get_arch).snap"
 }
 
+setup_snapd_proxy() {
+    if [ "${SNAPD_USE_PROXY:-}" != true ]; then
+        return
+    fi
+    restart=$1
+
+    mkdir -p /etc/systemd/system/snapd.service.d
+    cat <<EOF > /etc/systemd/system/snapd.service.d/proxy.conf
+[Service]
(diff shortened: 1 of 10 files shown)


tests/lib/prepare-utils.sh
```diff
@@ -158,6 +158,43 @@ get_core_snap_name() {
...
+    mkdir -p "$SNAPD_WORK_DIR"
+    cp -f /etc/environment "$SNAPD_WORK_DIR"/environment.bak
+    {
+        echo "HTTPS_PROXY=$HTTPS_PROXY"
+        echo "HTTP_PROXY=$HTTP_PROXY"
+        echo "https_proxy=$HTTPS_PROXY"
+        echo "http_proxy=$HTTP_PROXY"
+        echo "NO_PROXY=$NO_PROXY"
+        echo "no_proxy=$NO_PROXY"
+    } >> /etc/environment
```
valentindavid, line 195: `/etc/environment.d/*.conf` ?

valentindavid: Also I wonder if we could inject those another way, like kernel command line. There is systemd.setenv, but I am not sure it is the best.

Because things that  have started before we set that file will not work correctly. And I do not think it is nice to restart things.

valentindavid (approved):

PR #455, opened 2026-07-07
Makefile: enable changelog generation

Meulengracht:


tools/refresh-manifest.py
```diff
@@ -64,7 +64,7 @@ def _extract_record_path(record: dict) -> str | None:
 
     keys = PATH_KEYS_BY_KIND.get(kind)
     if not keys:
-        print(f'Unknown manifest record kind: {kind}', file=sys.stderr)
+        # likely a slice or something else that does not have a path, ignore it for now
```
Meulengracht, line 67: this path spams the logs incredibly much

Makefile
```diff
@@ -85,10 +85,31 @@ hooks:
 	mkdir -p "$(DESTDIR)/usr/share/snappy"
 	python3 ./tools/wall2dpkg.py "$(DESTDIR)/var/lib/chisel/manifest.wall" "$(DESTDIR)/usr/share/snappy/dpkg.yaml"
 
-	# TODO: Update the changelog generation to support chisel builds.
-
-	# TODO: Coordinate with the LP team that we now produce chisel artifacts
+	# When building through spread there is no .git, which means we cannot
```
Mohit-Chachada: looking at [this for core24](https://github.com/canonical/core-base/blob/core24/Makefile#L76-L80), shouldn't we make also sure that the current revision is updated to beta before generating the changelog (to get the diff w.r.t. beta)?

Makefile
```diff
@@ -85,10 +85,31 @@ hooks:
...
+	# generate the changelog in this case, ensure that the current folder is
+	# a git repository, additionally check that the snap we building a changelog
```
Mohit-Chachada:
```suggestion
	# a git repository, additionally check that the snap we are building a changelog
```

Makefile
```diff
@@ -85,10 +85,31 @@ hooks:
...
+		./tools/generate-changelog.py \
+			"/snap/$(SNAP_NAME)/current" \
+			"$(DESTDIR)" \
+			"$(SNAP_NAME)" \
+			$$CHG_PARAMS; \
+	else \
+		echo "WARNING: changelog will not be generated for this build"; \
+	fi
 
+	# only generate manifest and dpkg.yaml files for lp build
```
Mohit-Chachada: i suppose we don't have `usr/share/snappy/dpkg.list` for core26, so we don't copy it as the manifest, but we should update the above comment to reflect that
or probably we should copy `var/lib/chisel/manifest.wall` too to the lp build

Mohit-Chachada (commented):
one main question about refreshing to beta for changelog generation and a couple of minor points

Makefile
Meulengracht: It already is, it's just inside snapcraft.yaml;
```
    # to generate changelogs, it's own snap need to be present,
    # obviously for new core bases this cannot be the case, and
    # we need to wait until the first release with introducing
    # changelogs.
    build-snaps:
      - core26=latest/beta
```

Makefile
Meulengracht: ah drat, i missed this in the review of other snaps.

Makefile
Meulengracht: Good point about the manifest, updated.

Mohit-Chachada (approved):
Thanks!

Makefile
```diff
@@ -85,10 +85,32 @@ hooks:
...
+			"/snap/$(SNAP_NAME)/current" \
+			"$(DESTDIR)" \
+			"$(SNAP_NAME)" \
+			$$CHG_PARAMS; \
+	else \
+		echo "WARNING: changelog will not be generated for this build"; \
+	fi
 
+	# copy dpkg.yaml, chisel manifest and changelog files for lp build
+	if [ -e /build/$(SNAP_BUILD_NAME) ]; then \
```
valentindavid: Why do we want to not use what we generated, if we generated it? I see the changelog could have been generated on a local build, right?

Makefile
```diff
@@ -85,10 +85,32 @@ hooks:
...
+		/bin/cp $(DESTDIR)/usr/share/snappy/dpkg.yaml /build/$(SNAP_BUILD_NAME)/$(SNAP_NAME)-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).dpkg.yaml; \
+		/bin/cp $(DESTDIR)/var/lib/chisel/manifest.wall /build/$(SNAP_BUILD_NAME)/$(SNAP_NAME)-$$(date +%Y%m%d%H%M)_$(DPKG_ARCH).manifest.wall; \
+		if [ -e $(DESTDIR)/usr/share/doc/ChangeLog ]; then \
```
valentindavid: What are the cases where this is false?

Meulengracht: I think this was kind of a logically failure, because reflecting on this I added this as the changelog will not be generated for spread builds, but actually I'm not really sure it's even relevant as the initial `if` is not hit in spread afaict. I'll try to remove it and  see how it affects the spread build.

Makefile
Meulengracht: I'm confused by this comment, what do you mean? There is no LP build folder for other builds than on LP?

Meulengracht: The dpkg.yaml and changelog are still inside the snap? They are just not copied to the LP artifacts folder

valentindavid: Oh, that was confusing... Sorry. I thought this was run to build the snap.

valentindavid: We do call "make hooks" from snapcraft.yaml.

That is so weird thing. I wish snapcraft had a support for that.

Meulengracht: I agree with this, it's not really ideal

valentindavid: Could we move that to a separate step in snapcraft.yaml?

valentindavid: Also it would be nice to move what we can from changelog generation to a separate step as well.
And decide the CHG_PARAMS=--launchpad from snapcraft.yaml.

Meulengracht: Yea that we definitely could, we don't need anything from this specific step as we the artifacts are all located in CRAFT_PRIME, let me just test it out

valentindavid: For example wall2dpkg.py could be called separately also. But maybe not `refresh-manifest.py`. I do not know.

Meulengracht: Hmm good point, let me just test it out and see how it looks

Makefile
Meulengracht: The build works fine (not counting arm as it hit its quota), i'll keep this removed! Thank you for catching that

snapcraft.yaml
```diff
@@ -318,3 +311,57 @@ parts:
...
+        ./tools/generate-changelog.py "/snap/core26/current" "${CRAFT_PRIME}" core26 $CHG_PARAMS
+      else
+        echo "WARNING: changelog will not be generated for this build"
+      fi
+
+  publish-artifacts-lp:
+    after:
+      - generate-changelog
+    plugin: nil
+    source: .
```
valentindavid: Do we need sources  for this one?

snapcraft.yaml
```diff
@@ -318,3 +311,57 @@ parts:
 
       # ensure build-in tests are run
       cd ${CRAFT_PART_SRC} && make test TESTDIR=${CRAFT_PRIME}
+
+  generate-changelog:
+    after:
+      - bootstrap
+    plugin: make
+    source: .
```
valentindavid: Is it only things in ./tools we need here? Or we need anything else?

snapcraft.yaml
```diff
@@ -318,3 +311,57 @@ parts:
...
+      - python3-debian
+      - python3-requests
+      - python3-yaml
+    override-build: |
+      # no-op
+    override-prime: |
+      cd ${CRAFT_PART_SRC}
+
+      # to generate changelogs, it's own snap need to be present
+      snap refresh core26 --beta || snap install core26 --beta
```
valentindavid: What about?

```
build-snaps:
  - core26/beta
```

Meulengracht: yea was testing a few things, forgot to undo this change, thank you

snapcraft.yaml
Meulengracht: yea only tools, trying to reduce the source now, good point

snapcraft.yaml
Meulengracht: nope, will remove

snapcraft.yaml
```diff
@@ -318,3 +311,56 @@ parts:
...
+    source: tools
+    build-packages:
+      - python3-debian
+      - python3-requests
+      - python3-yaml
+    # to generate changelogs, it's own snap need to be present
+    build-snaps:
+      - core26/beta
+    override-build: |
+      # no-op
```
Mohit-Chachada: can we not get rid of this, if its a no-op?

snapcraft.yaml
```diff
@@ -282,20 +282,13 @@ parts:
     after:
       - chisel-libs
       - snapd-files
+      - generators
```
Mohit-Chachada: interesting, this doesn't quite seem to be caused due to the PR changes

Makefile
Mohit-Chachada: sorry to open this up again, but it seems that this could happen when `generate-changelog` is successful without actually generating the `usr/share/doc/ChangeLog`, i.e. for ex when `/snap/core26/current/usr/share/snappy/dpkg.yaml` is not available, which at least from my quick check in the core26 beta snap indeed shows that it is not present.
am i missing something?

snapcraft.yaml
Meulengracht: this was required for snapcraft building after the new recipe steps were introduced for some reason - i don't understand how snapcraft schedules the steps, but oddly enough nothing depended on them which i don't think was right anyway

snapcraft.yaml
Meulengracht: craftctl default invokes `make` because of the plugin type, but i cannot use nil or dump here

Makefile
Mohit-Chachada: this will work once an edge build is promoted to beta

snapcraft.yaml
```diff
@@ -323,3 +314,52 @@ parts:
...
+    build-snaps:
+      - core26/beta
+    override-build: |
+      # no-op
+    override-prime: |
+      cd ${CRAFT_PART_SRC}
+
+      # Generate the dpkg.yaml for compatability purposes. Scanning tools still expect
+      # this file to be present, and they move slowly in terms of support for the new chisel format.
+      mkdir -p "${CRAFT_PRIME}/usr/share/snappy"
```
Meulengracht: @valentindavid Ah, should I do this in build instead and then into _INSTALL

Meulengracht: Ah but we still run hooks in prime, then this is probably fine?

valentindavid: I think all this code could now be moved to override-build and install in CRAFT_PART_INSTALL after the refactoring I have done. I would really prefer that. It makes it easier for snapcraft to keep track of where files come from.

Meulengracht: The primary challenge is we need the manifest from the prime step in hooks, this step must not run before

Meulengracht: Alternatively i just move this entire step back into the Makefile, but that does not actually change anything

Meulengracht: As valentin said to me offline is that the makefile is now invoked in the normal build step of hooks.

valentindavid (approved):
Thank you!

PR #456, opened 2026-07-08
snapcraft.yaml: use snapcraft a bit more like it is intended

valentindavid:
We have been using prime too much and writing to stage (we should never write to stage, only read).

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: core26
-adopt-info: bootstrap
+adopt-info: hooks
 summary: Runtime environment based on Ubuntu 26.04
 description: |
   The base snap based on the Ubuntu 26.04 release.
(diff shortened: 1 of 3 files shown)


snapcraft.yaml
```diff
@@ -310,11 +317,48 @@ parts:
...
+      python3 ${CRAFT_PART_SRC}/wall2dpkg.py \
+        "${CRAFT_PART_INSTALL}/var/lib/chisel/manifest.wall" \
+        "${CRAFT_PART_INSTALL}/usr/share/snappy/dpkg.yaml"
+
+      # TODO: Update the changelog generation to support chisel builds.
+
+      # TODO: Coordinate with the LP team that we now produce chisel artifacts
+
+
+  test-step:
```
valentindavid, line 358: That was only to test that staging works. It has to be removed.

valentindavid:
Opened https://github.com/canonical/core-base/pull/457 instead which is a bit simpler first step.

PR #460, opened 2026-08-28
snapcraft: add variant-versioning for cloud-init and unify handling 

Meulengracht:
Introduce variant-versioning that covers cloud-init so versioning will look like:
```
20260629
20260629+cloud-init
20260629+fips
```
Based on the recipe naming instead of the git origin. This makes them easier to distinguish in  the store review tools as well, allowing hopefully for store reviews not to get hung up when there is a massive difference in snap contents.

Additionally this moves the step that generates the build-env to the front of the build, ensuring all other snapcraft tasks can use it.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -13,17 +13,9 @@ assumes:
   ## - isa-riscv64-rva23
 
 parts:
-  # TODO: rewrite this once all slices are upstreamed to use chisel slices
-  # directly from canonical, instead of our own chisel-releases repo.
-  chisel-libs:
+  build-env:
     plugin: nil
-    build-packages:
-      - golang
-      - file
-    source: https://github.com/canonical/chisel-ubuntu-core.git
-    source-type: git
-    source-branch: '26.04'
-    override-build: |
+    override-pull: |
       # We use a file with env vars in $CRAFT_PROJECT_DIR to decide whether to
       # include cloud-init or not. This file can be manually created for local
       # builds, or created here if we are running from a launchpad recipe with
@@ -50,6 +42,20 @@ parts:
             fi
           fi
       )
+
+  # TODO: rewrite this once all slices are upstreamed to use chisel slices
+  # directly from canonical, instead of our own chisel-releases repo.
+  chisel-libs:
+    plugin: nil
+    after:
+      - build-env
+    build-packages:
+      - golang
+      - file
+    source: https://github.com/canonical/chisel-ubuntu-core.git
+    source-type: git
+    source-branch: '26.04'
+    override-build: |
       go install github.com/canonical/chisel/cmd/chisel@latest
       
       # Define the system utilities that we want to include in the base
@@ -294,6 +300,7 @@ parts:
       - chisel-libs
       - snapd-files
       - splash-theme
+      - build-env
     plugin: make
     source: .
     build-packages:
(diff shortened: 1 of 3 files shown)


snapcraft.yaml
```diff
@@ -41,10 +33,29 @@ parts:
       (
           cd "$CRAFT_PROJECT_DIR"
           touch build-env
-          if printf "$SNAPCRAFT_IMAGE_INFO" | grep -q -- "-cloud-init/+build"
-          then echo "SNAP_BUILD_VARIANT=cloud-init" > build-env
+          # parse the snap name from the url, and remove the prefix "core26-" and 
+          # suffix "/+build" to get the variant name, if any.
+          if [ -n "${SNAPCRAFT_IMAGE_INFO+x}" ]; then
```
valentindavid: Is there a way to make this run again if `SNAPCRAFT_IMAGE_INFO` changed? It looks like there is no source for this part. So I expect it is not going to re-run if changed.

snapcraft.yaml
```diff
@@ -41,10 +33,29 @@ parts:
...
+            SNAP_BUILD_VARIANT=$(printf '%s' "$SNAPCRAFT_IMAGE_INFO" | grep -oP '(?<=core26-).*(?=/\+build)')
+            if [ -n "$SNAP_BUILD_VARIANT" ]; then
+                echo "SNAP_BUILD_VARIANT=$SNAP_BUILD_VARIANT" > build-env
```
valentindavid: override-pull should write into `${CRAFT_PART_SRC}`. Then you should install that file in build in CRAFT_PART_INSTALL. And remove it from prime.

Otherwise, if you re-run that part and the SNAP_BUILD_VARIANT is empty, the `build-env` file will not be removed.

snapcraft.yaml
Meulengracht: SNAPCRAFT_IMAGE_INFO is only set by LP, and we do not re-run builds there. This seems unnecessary.

snapcraft.yaml
Meulengracht: This has been updated after some in-depth explanation of the parts lifecycle on MM.

snapcraft.yaml
```diff
@@ -38,13 +31,55 @@ parts:
...
-      (
-          cd "$CRAFT_PROJECT_DIR"
-          touch build-env
-          if printf "$SNAPCRAFT_IMAGE_INFO" | grep -q -- "-cloud-init/+build"
-          then echo "SNAP_BUILD_VARIANT=cloud-init" > build-env
-          fi
-      )
+      if ! [ -e "$CRAFT_PROJECT_DIR"/build-env ]; then
+        (
+            cd "${CRAFT_PART_INSTALL}"
```
valentindavid: override-pull should not write in CRAFT_PART_INSTALL, but in CRAFT_PART_SRC

valentindavid: You can use override-build to copy then.

snapcraft.yaml
```diff
@@ -13,17 +13,10 @@ assumes:
...
+  build-env:
     plugin: nil
-    build-packages:
-      - golang
-      - file
-    source: https://github.com/canonical/chisel-ubuntu-core.git
-    source-type: git
-    source-branch: '26.04'
-    override-build: |
+    source: .
```
valentindavid, line 19: Maybe comment that it is to retrigger pull automatically, but we do not call `craftctl default` in `override-pull`.

Meulengracht: added a comment

valentindavid (approved):

PR #461, opened 2026-09-01
snapcraft: include nsenter and pivot_root

Meulengracht:
This will include:
- util-linux_nsenter
- util-linux-extra_switch-root

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -120,6 +120,8 @@ parts:
         squashfs-tools_bins
         tar_bins
         util-linux_bins
+        util-linux_nsenter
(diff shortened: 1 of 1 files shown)


bboozzoo (approved):
Thanks!

valentindavid (commented):
The only binaries that both exists in core24 AND are allow by some apparmor profile from interfaces (or base template) are nsenter and pivot_root. So it seems the others are maybe not needed.

valentindavid:
* network_control needs nsenter
* greengrass_support needs pivot_root

valentindavid (approved):

PR #464, opened 2026-09-02
github: integrate cve-scanning as a part of the regular CI

Meulengracht:
Ported from #462 for core24

Unfortunately, to make this even run we have to migrate to openstack and fix some pretty fundamental things here as the pipeline was completely broken


valentindavid (approved):
We need to take care of NTP on openstack in follow up.

